PR #15, opened 2023-02-03
Experimental Lean4 backend

msprotz:
This PR aims to merge early support for a Lean4 backend.

Right now, the backend generates syntactically-correct code, at least for the hashmap example. The code can be tested by running `lake build` in tests/lean/hashmap_on_disk, followed by `build/bin/hashmap` ; one can also open the code in vscode and step through the various #eval statements.

There are still plenty of things to improve, obviously, but let's get started with the review process and polish things while it's still early work.


tests/lean/hashmap_on_disk/HashmapMain/Clauses/Template.lean
```diff
@@ -0,0 +1,113 @@
+-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
```
sonmarcho, line 1: We should have a `Template.lean` generated by Aeneas, and a `Clauses.lean` which is user written and opened by the other files

tests/lean/hashmap_on_disk/HashmapMain/Funs.lean
```diff
@@ -0,0 +1,658 @@
+-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
+-- [hashmap_main]: function definitions
+import Base.Primitives
+import HashmapMain.Types
+import HashmapMain.Opaque
+import HashmapMain.Clauses.Template
```
sonmarcho, line 6: We shouldn't import the `Template` file directly, but a user-written `Clauses` file (we generate the `Template` file just to help the user). I noticed this also changed for the other backends.

tests/lean/hashmap_on_disk/HashmapMain/Funs.lean
```diff
@@ -0,0 +1,658 @@
...
+  if n > (USize.ofNatCore 0 (by intlit))
+  then
+    do
+      let slots0 <-
+        vec_push_back (hashmap_list_t T) slots hashmap_list_t.HashmapListNil 
+      let n0 <- USize.checked_sub n (USize.ofNatCore 1 (by intlit)) 
+      hashmap_hash_map_allocate_slots_loop_fwd T slots0 n0
+  else result.ret slots
+termination_by hashmap_hash_map_allocate_slots_loop_fwd slots n =>
+                                                                  hashmap_hash_map_allocate_slots_loop_terminates
```
sonmarcho, line 27: I think the indentation should be a simple fix

tests/lean/hashmap_on_disk/HashmapMain/Funs.lean
```diff
@@ -0,0 +1,658 @@
...
+def hashmap_hash_map_contains_key_in_list_loop_fwd
+  (T : Type) (key : USize) (ls : hashmap_list_t T) : (result Bool) :=
+  match ls with
+  | hashmap_list_t.HashmapListCons ckey t tl =>
+    if ckey = key
+    then result.ret true
+    else hashmap_hash_map_contains_key_in_list_loop_fwd T key tl
+  | hashmap_list_t.HashmapListNil => result.ret false
+  
+termination_by hashmap_hash_map_contains_key_in_list_loop_fwd key ls =>
```
sonmarcho, line 318: I'm surprised there is a space above

tests/lean/hashmap_on_disk/HashmapMain/Types.lean
```diff
@@ -0,0 +1,19 @@
...
+
+/- [hashmap_main::hashmap::List] -/
+inductive hashmap_list_t (T : Type) :=
+| HashmapListCons : USize -> T -> hashmap_list_t T -> hashmap_list_t T
+| HashmapListNil : hashmap_list_t T
+
+/- [hashmap_main::hashmap::HashMap] -/
+structure hashmap_hash_map_t (T : Type) where
+
+  hashmap_hash_map_num_entries : USize
```
sonmarcho, line 13: I'm surprised there is a space above

tests/lean/hashmap_on_disk/Main.lean
```diff
@@ -0,0 +1,10 @@
+import Base.Primitives
```
sonmarcho, line 1: This file is hand-written, correct?

tests/lean/hashmap_on_disk/lakefile.lean
```diff
@@ -0,0 +1,22 @@
+import Lake
```
sonmarcho, line 1: I guess we should generate the `lakefile.lean` files

tests/lean/hashmap_on_disk/lakefile.lean
```diff
@@ -0,0 +1,22 @@
...
+lean_lib «HashmapMain» {
+  -- add library configuration options here
+}
+
+lean_lib «Base» {
+  -- add library configuration options here
+}
+
+@[default_target]
+lean_exe «hashmap» {
```
sonmarcho, line 20: Do we really need a lean_exe?

tests/lean/hashmap_on_disk/lakefile.lean
sonmarcho: I can do it, but I'm not sure what we should put in the various fields

compiler/ExtractBase.ml
```diff
@@ -486,6 +509,19 @@ let id_to_string (id : id) (ctx : extraction_ctx) : string =
         | Some lid -> ", loop: " ^ LoopId.to_string lid
       in
       "decreases clause for function: " ^ fun_name ^ loop
+  | TerminatesClauseId (fid, lid) ->
```
sonmarcho, line 512: I think we should change the names: I see F\* uses the name `decreases` while Lean uses the name `terminates`.
What annoys me is that depending on the backend `DecreasesClause` stands either for the decreasing measure (F\*) or for a tactic (Lean). I think we should use the name `DecreasesClause` and `DecreasesTac`, or `TerminatesClause` and `TerminatesTac`.

compiler/ExtractBase.ml
```diff
@@ -279,17 +294,24 @@ type formatter = {
 type id =
   | GlobalId of A.GlobalDeclId.id
   | FunId of fun_id
+  | DeclaredId of fun_id
```
sonmarcho, line 297: We shouldn't need this: declared ids should be of kind `FunId`

compiler/ExtractBase.ml
```diff
@@ -730,6 +781,13 @@ let ctx_add_fun_decl (trans_group : bool * pure_fun_translation)
     ctx.fmt.fun_name def.basename def.num_loops def.loop_id num_rgs rg_info
       (keep_fwd, num_backs)
   in
+  let ctx = if def.body = None && !Config.backend = Lean then
```
sonmarcho, line 784: We shouldn't have this

compiler/Extract.ml
```diff
@@ -120,16 +132,18 @@ let keywords () =
           "tt";
           "char_of_byte";
         ]
+    | Lean ->
+        [] (* TODO *)
```
sonmarcho, line 136: TODO: start filling this (it's not that important, but useful)

compiler/Extract.ml
```diff
@@ -191,15 +215,24 @@ let assumed_pure_functions : (pure_assumed_fun_id * string) list =
   | Coq ->
       (* We don't provide [FuelDecrease] and [FuelEqZero] on purpose *)
       [ (Return, "return_"); (Fail, "fail_"); (Assert, "massert") ]
+  | Lean ->
+      [
+        (Return, "return");
+        (Fail, "fail_");
+        (Assert, "massert");
+        (* TODO: figure out how to deal with this *)
```
sonmarcho, line 223: We can omit those

sonmarcho (commented):
This overall looks good, and it was a lot of work :)

I put many remarks for minor improvements/fixes that I would like to see: I can take care of most of them.

Those put aside, there are 2 things we need to add:
- we should generate Lean files for the whole test suite (not juste the hashmap): this should be easy, put aside the generation of the `lakefile.lean` file for the `misc` tests (but maybe for lean we should simply generate the code in subfolders for `misc`)
- we should set up Nix for Lean

tests/lean/hashmap_on_disk/HashmapMain/Clauses/Template.lean
sonmarcho: I fixed this: we now need to remove the `sorry` from the `Clauses.lean` files

tests/lean/hashmap_on_disk/HashmapMain.lean
```diff
@@ -0,0 +1 @@
+def hello := "world"
```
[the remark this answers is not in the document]
msprotz: this was added by `lake init` as a placeholder... removed it

tests/lean/hashmap_on_disk/HashmapMain/Clauses/Template.lean
msprotz: Clauses.lean was missing so I added it for the sake of restoring testing via `lake build` -- agreed that we need to kill the missing proofs in there, but first we need to add reasoning principles for standard library functions, such as checked_sub, and so on

tests/lean/hashmap_on_disk/HashmapMain/Funs.lean
msprotz: I'm resolving this on the basis that your comment above says that you fixed it (and indeed the code-gen seems to indicate that this is fixed... thanks for fixing it, I think I misunderstood the intended usage, or maybe didn't realize what was going on at first)

tests/lean/hashmap_on_disk/Main.lean
msprotz: correct ... this one I'm leaving in here for the sake of unit-testing, until we have proper support for #assert -- right now, running the code does not check the #asserts yet

tests/lean/hashmap_on_disk/lakefile.lean
msprotz: right now we need it for the sake of running the tests, since I have yet to find a way to implement #assert in a way that it runs #eval then checks for `True` -- I find it convenient to have a placeholder, for a larger example, this would be a good place to add e.g. more unit-tests, performance benchmarks, etc.

tests/lean/hashmap_on_disk/lakefile.lean
msprotz: basically everything I have so far is (I believe) useful -- just replace the occurrences of `hashmap` with `<name-of-the-current-file>`

you probably want to keep the mathlib import, I currently need it for `run_cmd`, it's a little heavy-handed, but convenient

backends/lean/Primitives.lean
```diff
@@ -0,0 +1,373 @@
...
+-- NOTE: we reuse the USize type from prelude.lean, because at least we know
+-- it's defined in an idiomatic style that is going to make proofs easy (and
+-- indeed, several proofs here are much shortened compared to Aymeric's earlier
+-- attempt.) This is not stricto sensu the *correct* thing to do, because one
+-- can query at run-time the value of USize, which we do *not* want to do (we
+-- don't know what target we'll run on!), but when the day comes, we'll just
+-- define our own USize.
+-- ANOTHER NOTE: there is USize.sub but it has wraparound semantics, which is
+-- not something we want to define (I think), so we use our own monadic sub (but
+-- is it in line with the Rust behavior?)
```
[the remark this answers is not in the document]
msprotz: Fixed all of the comments in that area

backends/lean/Primitives.lean
```diff
@@ -0,0 +1,373 @@
...
+-- Let-binding the result of a monadic operation is oftentimes not sufficient,
+-- because we may need a hypothesis for equational reasoning in the scope. We
+-- rely on subtype, and a custom let-binding operator, in effect recreating our
+-- own variant of the do-dsl
+
+def result.attach : (o : result α) → result { x : α // o = ret x }
+  | .ret x => .ret ⟨x, rfl⟩
+  | .fail e   => .fail e
+
+macro "let" h:ident " : " e:term " <-- " f:term : doElem =>
```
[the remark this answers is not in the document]
msprotz: Added a comment to remember

backends/lean/Primitives.lean
```diff
@@ -0,0 +1,373 @@
...
+-- MACHINE INTEGERS --
+----------------------
+
+-- NOTE: we reuse the USize type from prelude.lean, because at least we know
+-- it's defined in an idiomatic style that is going to make proofs easy (and
+-- indeed, several proofs here are much shortened compared to Aymeric's earlier
+-- attempt.) This is not stricto sensu the *correct* thing to do, because one
+-- can query at run-time the value of USize, which we do *not* want to do (we
+-- don't know what target we'll run on!), but when the day comes, we'll just
+-- define our own USize.
```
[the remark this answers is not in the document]
msprotz: Added a comment, along the lines of "16-bit is a niche use-case, we'll deal with it later"

backends/lean/Primitives.lean
```diff
@@ -0,0 +1,373 @@
...
+    .fail integerOverflow
+
+
+-- Test behavior...
+#eval assert! USize.checked_sub 10 20 == fail integerOverflow; 0
+
+#eval USize.checked_sub 20 10
+-- NOTE: compare with concrete behavior here, which I do not think we want
+#eval USize.sub 0 1
+#eval UInt8.add 255 255
```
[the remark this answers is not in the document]
msprotz: Added a comment, we need to figure out #assert first

compiler/ExtractBase.ml
msprotz: I picked the names for consistency with the syntax, not the semantics:
- an F* termination *measure* (introduced by `decreases e`) is mapped via a DecreasesClauseId
- a Lean termination *measure* (introduced by `termination_by f x y z => e`) is mapped via a TerminatesClauseId
- a Lean termination *proof* (introduced by `decreasing_by <<tactic>>`) is mapped via a DecreasesClauseId 

Remember that in Lean4, the syntax is

```lean
termination_by f x y z => (x, y, z)
decreasing_by
  simp_wf
```

If you're arguing for a more semantic approach to this mapping, I would say we pick `TerminationMeasure` and `DecreasingProof` and map the former to `decreases` (F*) or `termination_by` (Lean), and the latter to `decreasing_by` (Lean only).

Thoughts?

compiler/ExtractBase.ml
msprotz: Yeah my problem was that I needed a name for the definition, and a name for the usage (prefixed with `OpaqueDefs`), hence the two kinds of entries. If you know how to fix this cleanly, please go ahead? I'm a little confused by the handling of names right now

compiler/ExtractBase.ml
msprotz: What would you like instead?

compiler/SymbolicToPure.ml
```diff
@@ -2615,6 +2615,8 @@ let wrap_in_match_fuel (fuel0 : VarId.id) (fuel : VarId.id) (body : texpression)
       let match_ty = body.ty in
       let match_e = Switch (fuel0, Match [ fail_branch; success_branch ]) in
       { e = match_e; ty = match_ty }
+  | Lean ->
+      failwith "Not handling fuel in Lean"
```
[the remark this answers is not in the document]
msprotz: What kind of check?

compiler/Translate.ml
```diff
@@ -558,7 +563,10 @@ let export_functions_group (fmt : Format.formatter) (config : gen_config)
         let extract_decrease decl =
           let has_decr_clause = has_decreases_clause decl in
           if has_decr_clause then
-            Extract.extract_template_decreases_clause ctx.extract_ctx fmt decl
+            if !Config.backend = Lean then
+              Extract.extract_termination_and_decreasing ctx.extract_ctx fmt decl
+            else
+              Extract.extract_template_decreases_clause ctx.extract_ctx fmt decl
```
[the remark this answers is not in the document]
msprotz: The two syntaxes were so wildly different that I thought it would obfuscate things more to try to share code between them.

compiler/Translate.ml
```diff
@@ -653,13 +661,24 @@ let extract_definitions (fmt : Format.formatter) (config : gen_config)
    *)
   if config.extract_state_type && config.extract_fun_decls then
     export_state_type ();
+  if config.extract_opaque && config.extract_fun_decls && !Config.backend = Lean then begin
+    Format.pp_print_break fmt 0 0;
+    Format.pp_open_vbox fmt ctx.extract_ctx.indent_incr;
+    Format.pp_print_string fmt "structure OpaqueDefs where";
+    Format.pp_print_break fmt 0 0
+  end;
```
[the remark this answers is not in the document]
msprotz: Added a comment, but this doesn't work in single-file mode

tests/lean/hashmap_on_disk/lakefile.lean
sonmarcho: I guess that for the time being, whenever there are unit tests, in Lean, we could generate a `Main` file which groups all the unit tests, and declare it as `lean_exe` ?

compiler/ExtractBase.ml
sonmarcho: Following a 1:1 discussion : we will remove `DeclaredId` and register the name with the prefix `OpaqueDefs.`

sonmarcho:
Following a 1:1 discussion: we should print a warning that the Lean backend is experimental whenever the user chooses this warning (and print a warning about loops, if the crate to translate contains loops).

compiler/Translate.ml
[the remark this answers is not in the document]
sonmarcho: I rechecked this: indeed, the syntax is indeed quite different. As a consequence, I renamed the functions to make it clear they are specific to Lean and F*.

compiler/ExtractBase.ml
sonmarcho: Finally opted for the following:
- we register the definition names without any prefix, but remember which definitions are opaque
- at lookup time, depending on the usage, we add or not an "opaque_defs." prefix to the opaque definition names

compiler/Translate.ml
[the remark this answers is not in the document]
sonmarcho: For now I check that the files are split if we use opaque functions and target the Lean backend.
If we want to make this general, we will need to parameterize the implementation with two modules: `OpaqueTypes` and `OpaqueFuns`, and the declarations in `OpaqueFuns` will be able to refer to the declarations in `OpaqueTypes`.

tests/lean/hashmap_on_disk/lakefile.lean
sonmarcho: I automated the generation of the lakefile.lean files.
For now:
- the package name is the crate name in snake case
- there is no default target

tests/lean/hashmap_on_disk/Main.lean
sonmarcho: As I understand that the `#assert` now works, I removed this file

sonmarcho:
I took care of all the points listed above at the exception of:
- update the Makefiles to verify the generated Lean files
- setup Nix

sonmarcho:
Setting up a Nix derivation for Lean is a lot less easy than I thought: after discussion I will merge this PR and add a Nix derivation in a subsequent PR. Also, there are still many things to improve with the current version of the backend, so it is as of today largely experimental.

PR #39, opened 2023-09-22
Add support for bitshifts

R1kM:
Adds Lean support for bitshifts, using the `>>>` and `<<<` operators.
Note, compared to other binary operators, the Rust compiler accepts operands of different integer types for bitshifts.


sonmarcho:
I updated the extraction to make it work, and added support for the other bitwise operators (and, or, xor). I left several admits for the definitions in the backends, which I leave for a future PR.

PR #106, opened 2024-04-03
Added Error and EError to expressions and propagated related changes

EschericHya:
Work on error catching

### DIFF
--- a/compiler/ExtractBase.ml
+++ b/compiler/ExtractBase.ml
@@ -1701,7 +1701,7 @@ let ctx_compute_var_basename (meta : Meta.meta) (ctx : extraction_ctx)
           match lty with TBool -> "b" | TChar -> "c" | TInteger _ -> "i")
       | TArrow _ -> "f"
       | TTraitType (_, name) -> name_from_type_ident name
-      | Error -> "@Error")
+      | Error -> "x")
 (* TODO : Check*)
 
 (** Generates a type variable basename. *)
--- a/compiler/ExtractBase.ml
+++ b/compiler/ExtractBase.ml
@@ -259,7 +259,7 @@ let report_name_collision (id_to_string : id -> string)
   let meta_to_string (meta : Meta.meta option) =
     match meta with
     | None -> ""
-    | Some meta -> "\n  " ^ Errors.meta_to_string meta.span
+    | Some meta -> "\n  " ^ Errors.meta_to_string meta
   in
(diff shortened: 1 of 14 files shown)


compiler/ExtractBase.ml
```diff
@@ -1700,7 +1700,9 @@ let ctx_compute_var_basename (meta : Meta.meta) (ctx : extraction_ctx)
       | TLiteral lty -> (
           match lty with TBool -> "b" | TChar -> "c" | TInteger _ -> "i")
       | TArrow _ -> "f"
-      | TTraitType (_, name) -> name_from_type_ident name)
+      | TTraitType (_, name) -> name_from_type_ident name
+      | Error -> "@Error")
+(* TODO : Check*)
```
sonmarcho, line 1705: Use "x"

compiler/PrintPure.ml
```diff
@@ -615,6 +616,9 @@ let rec texpression_to_string ?(metadata : Meta.meta option = None)
           let e = meta_s ^ "\n" ^ indent ^ e in
           if inside then "(" ^ e ^ ")" else e
       | MPlace _ -> "(" ^ meta_s ^ " " ^ e ^ ")")
+  | EError (meta, msg) ->
+      if Option.is_none meta then msg
+      else meta_to_string (Option.get meta) ^ " " ^ msg (* TODO formatting *)
```
sonmarcho, line 621: I think it's ok to just have "@ ERROR" (without the space)

compiler/SymbolicToPure.ml
```diff
@@ -1963,6 +1963,9 @@ let eval_ctx_to_symbolic_assignments_info (ctx : bs_ctx)
   (* Return the computed information *)
   !info
 
+let translate_meta (meta : Meta.meta option) (msg : string) : texpression =
+  { e = EError (meta, msg); ty = Error }
```
sonmarcho, line 1967: This shouldn't be named `translate_meta` but `translate_error`

sonmarcho (changes requested):

compiler/ExtractBase.ml
sonmarcho: Could you remove the "TODO" comment?...

sonmarcho:
Could you add some code in Main.ml which checks whether some errors were registered and prints all the registered errors before exiting with an error code?
You can update this `try catch` to do nothing (note that if we catch an error, then this error must have been registered, so we don't need to register it again) and after the `try catch` you can check the list of errors.

PR #110, opened 2024-04-04
Add builtins for some checked ops for the Lean backend

zhassan-aws:
Extend Aeneas's Lean library with checked operations on scalars (e.g. `checked_add`, `checked_sub`, etc.).


backends/lean/Base/Primitives/Scalar.lean
```diff
@@ -568,6 +568,35 @@ instance {ty} : HOr (Scalar ty) (Scalar ty) (Scalar ty) where
 instance {ty} : HAnd (Scalar ty) (Scalar ty) (Scalar ty) where
   hAnd x y := Scalar.and x y
 
+-- core checked arithmetic operations
+
+/- A helper function that converts failure to none and success to some
+   TODO: move up to Base module? -/
+@[simp] abbrev Option.ofResult : Result (Scalar ty) → Result (Option (Scalar ty))
```
zhassan-aws, line 575: I wonder if this should just return `Option` instead of `Result Option`? Not sure what to do about the `div` cause though.

sonmarcho: I think the `Result` it takes as input should not be the same as the `Result` in the output: the input is the Rust result type (https://doc.rust-lang.org/std/result/enum.Result.html) while the output result is our way of modelling effects such as panics and divergence. I can take care of that.

sonmarcho: Yes, that's a good remark. I indeed think we can anticipate on the fact that we will remove the result type in some places to not have a `Result` here. Also, for all the cases below we can't have the `div` case, so that's fine.
I'm also starting to think that maybe we should define the base operations with an `Option`, then convert the option to a `Result` so that here you don't have to go the other way around.
Let me make the changes from here (thanks for the PR!).

zhassan-aws:
> I guess you're fine if I push changes directly into your branch?

Of course! Let me know if I need to update its permissions.

sonmarcho:
I substantially modified the definitions. However, I added lemmas like `core.num.checked_add_spec` which should give your what you need (you might have to state and prove your own auxiliary lemmas, but it should be easy to prove them by using the `core.num.checked_*_spec` theorems).

PR #113, opened 2024-04-05
Error catching should tell when code couldn't be generated

EschericHya:
Error catching should now be able to tell when code couldn't be generated

### DIFF
--- a/compiler/Extract.ml
+++ b/compiler/Extract.ml
@@ -1879,102 +1879,104 @@ let extract_global_decl_hol4_opaque (meta : Meta.meta) (ctx : extraction_ctx)
     [{start,end}_gloabl_decl_group], contrary to {!extract_type_decl}
     and {!extract_fun_decl}.
  *)
-let extract_global_decl (ctx : extraction_ctx) (fmt : F.formatter)
-    (global : global_decl option) (body : fun_decl) (interface : bool) : unit =
+let extract_global_decl_aux (ctx : extraction_ctx) (fmt : F.formatter)
+    (global : global_decl) (body : fun_decl) (interface : bool) : unit =
   let meta = body.meta in
   sanity_check __FILE__ __LINE__ body.is_global_decl_body meta;
   sanity_check __FILE__ __LINE__ (body.signature.inputs = []) meta;
-  match global with
-  | Some global -> (
-      (* Add a break then the name of the corresponding LLBC declaration *)
-      F.pp_print_break fmt 0 0;
-      let name =
-        if !Config.extract_external_name_patterns && not global.is_local then
-          Some global.llbc_name
-        else None
-      in
-      extract_comment_with_span ctx fmt
-        [ "[" ^ name_to_string ctx global.llbc_name ^ "]" ]
-        name global.meta.span;
-      F.pp_print_space fmt ();
+  (* Add a break then the name of the corresponding LLBC declaration *)
+  F.pp_print_break fmt 0 0;
+  let name =
+    if !Config.extract_external_name_patterns && not global.is_local then
+      Some global.llbc_name
+    else None
+  in
+  extract_comment_with_span ctx fmt
+    [ "[" ^ name_to_string ctx global.llbc_name ^ "]" ]
+    name global.meta.span;
+  F.pp_print_space fmt ();
 
-      let decl_name = ctx_get_global meta global.def_id ctx in
-      let body_name =
-        ctx_get_function meta
-          (FromLlbc (Pure.FunId (FRegular global.body_id), None))
-          ctx
-      in
-      let decl_ty, body_ty =
-        let ty = body.signature.output in
-        if body.signature.fwd_info.effect_info.can_fail then
-          (unwrap_result_ty meta ty, ty)
-        else (ty, mk_result_ty ty)
-      in
-      (* Add the type parameters *)
-      let ctx, type_params, cg_params, trait_clauses =
-        ctx_add_generic_params meta global.llbc_name global.llbc_generics
-          global.generics ctx
-      in
-      match body.body with
-      | None ->
-          (* No body: only generate a [val x_c : u32] declaration *)
-          let kind = if interface then Declared else Assumed in
-          if !backend = HOL4 then
-            extract_global_decl_hol4_opaque meta ctx fmt decl_name
-              global.generics decl_ty
-          else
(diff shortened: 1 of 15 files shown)


compiler/Extract.ml
```diff
@@ -1880,99 +1880,102 @@ let extract_global_decl_hol4_opaque (meta : Meta.meta) (ctx : extraction_ctx)
     and {!extract_fun_decl}.
  *)
 let extract_global_decl (ctx : extraction_ctx) (fmt : F.formatter)
-    (global : global_decl) (body : fun_decl) (interface : bool) : unit =
+    (global : global_decl option) (body : fun_decl) (interface : bool) : unit =
```
sonmarcho, line 1883: This creates a huge diff for nothing.
Could you have an auxiliary `extract_global_decl_aux` instead (which is the original `extract_global_decl`) and an `extract_global_decl` which has an optional global decl parameter instead?

compiler/Translate.ml
```diff
@@ -41,7 +41,7 @@ let translate_function_to_symbolics (trans_ctx : trans_ctx) (fdef : fun_decl) :
     of backward functions, we also provide names for the outputs.
     TODO: maybe we should introduce a record for this.
 *)
-let translate_function_to_pure (trans_ctx : trans_ctx)
+let translate_function_to_pure_hook (trans_ctx : trans_ctx)
```
sonmarcho, line 44: I prefer the convention "_aux" (this is not a "hook")

compiler/Translate.ml
```diff
@@ -195,6 +195,17 @@ let translate_function_to_pure (trans_ctx : trans_ctx)
...
 
+let translate_function_to_pure (trans_ctx : trans_ctx)
+    (pure_type_decls : Pure.type_decl Pure.TypeDeclId.Map.t)
+    (fun_dsigs : Pure.decomposed_fun_sig FunDeclId.Map.t) (fdef : fun_decl) :
+    pure_fun_translation_no_loops option =
+  try
+    Some
+      (translate_function_to_pure_hook trans_ctx pure_type_decls fun_dsigs fdef)
+  with CFailure (meta, _) ->
+    let () = save_error __FILE__ __LINE__ meta "Could not generate code, see previous error" in
```
sonmarcho, line 206: The comment should be something along the lines of "Could not translate function ... because of previous error"

compiler/Translate.ml
```diff
@@ -220,32 +231,51 @@ let translate_crate_to_pure (crate : crate) :
...
-             SymbolicToPure.translate_fun_sig_from_decl_to_decomposed trans_ctx
-               fdef ))
+           try
+             Some
+               ( fdef.def_id,
+                 SymbolicToPure.translate_fun_sig_from_decl_to_decomposed
+                   trans_ctx fdef )
+           with CFailure (meta, _) ->
+             let () =
+               save_error __FILE__ __LINE__ meta "Could not generate code, see previous error"
```
sonmarcho, line 243: Same remark for the comment as for the functions

compiler/Translate.ml
```diff
@@ -220,32 +231,51 @@ let translate_crate_to_pure (crate : crate) :
...
   (* Translate the trait declarations *)
   let trait_decls =
-    List.map
-      (SymbolicToPure.translate_trait_decl trans_ctx)
+    List.filter_map
+      (fun a ->
+        try Some (SymbolicToPure.translate_trait_decl trans_ctx a)
+        with CFailure (meta, _) ->
+          let () =
+            save_error __FILE__ __LINE__ meta "Could not generate code, see previous error"
```
sonmarcho, line 263: Same remark as above

compiler/Translate.ml
```diff
@@ -220,32 +231,51 @@ let translate_crate_to_pure (crate : crate) :
...
   (* Translate the trait implementations *)
   let trait_impls =
-    List.map
-      (SymbolicToPure.translate_trait_impl trans_ctx)
+    List.filter_map
+      (fun a ->
+        try Some (SymbolicToPure.translate_trait_impl trans_ctx a)
+        with CFailure (meta, _) ->
+          let () =
+            save_error __FILE__ __LINE__ meta "Could not generate code, see previous error"
```
sonmarcho, line 276: Same remark as above

compiler/Translate.ml
```diff
@@ -471,7 +501,12 @@ let export_global (fmt : Format.formatter) (config : gen_config) (ctx : gen_ctx)
        groups are always singletons, so the [extract_global_decl] function
        takes care of generating the delimiters.
     *)
-    let global = SymbolicToPure.translate_global ctx.trans_ctx global in
+    let global =
+      try Some (SymbolicToPure.translate_global ctx.trans_ctx global)
+      with CFailure (meta, _) ->
+        let () = save_error __FILE__ __LINE__ meta "Could not generate code, see previous error" in
```
sonmarcho, line 507: Same remark as above

sonmarcho (changes requested):

compiler/Translate.ml
```diff
@@ -220,32 +234,54 @@ let translate_crate_to_pure (crate : crate) :
...
-               fdef ))
+           try
+             Some
+               ( fdef.def_id,
+                 SymbolicToPure.translate_fun_sig_from_decl_to_decomposed
+                   trans_ctx fdef )
+           with CFailure (meta, _) ->
+             let () =
+               save_error __FILE__ __LINE__ meta
+                 "Could not translate function because of previous error"
```
sonmarcho, line 247: It is a function signature, not a function

compiler/SymbolicToPure.ml
```diff
@@ -3861,7 +3861,15 @@ let translate_fun_decl (ctx : bs_ctx) (body : S.expression option) : fun_decl =
...
 
 let translate_type_decls (ctx : Contexts.decls_ctx) : type_decl list =
-  List.map (translate_type_decl ctx)
+  List.filter_map
+    (fun a ->
+      try Some (translate_type_decl ctx a)
+      with CFailure (meta, _) ->
+        let () =
+          save_error __FILE__ __LINE__ meta
+            "Could not generate code, see previous error"
```
sonmarcho, line 3870: Error message

sonmarcho (changes requested):
There are still some problematic comments.
Also, could you add the name of the declaration we don't manage to extract inside the error message?

PR #115, opened 2024-04-08
Update flake.nix

Nadrieril:
It was missing some dependencies, and while we're at it we can disable sanity checks in tests.


sonmarcho (approved):
This is a good idea, thanks for this.

Issue #118, opened 2024-04-10
Support non-primitive `Copy` types


Nadrieril:
This code errors with "Not primitively copyable":
```rust
fn test<T: Copy>(x: T) {
    let y = x;
    let z = x;
}
```
This is because charon emits `copy(..)` operands without extra trait information. Given that rustc already checks that the types are `Copy`, we can probably ignore `copy(..)` entirely on the aeneas side of things.

PR #119, opened 2024-04-10
Trust rustc regarding `Copy` bounds

Nadrieril:
Rustc's type checking guarantees that any `copy(..)` in LLBC applies to a type that has the `Copy` bound. Because `copy(..)` means "copying just the bits", there is no need to find the corresponding `Copy` implementation either, so https://github.com/AeneasVerif/charon/pull/111 makes `ty_is_copyable` always return `true`.

This bumps charon to https://github.com/AeneasVerif/charon/pull/111.

Fixes https://github.com/AeneasVerif/aeneas/issues/118

### DIFF
--- a/compiler/Invariants.ml
+++ b/compiler/Invariants.ml
@@ -825,7 +825,7 @@ let check_symbolic_values (meta : Meta.meta) (ctx : eval_ctx) : unit =
      * it must be expanded first *)
     if ty_has_borrows ctx.type_ctx.type_infos info.ty then
       sanity_check __FILE__ __LINE__ (info.env_count <= 1) meta;
-    (* A duplicated symbolic value is necessarily primitively copyable *)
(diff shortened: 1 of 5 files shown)


compiler/Invariants.ml
```diff
@@ -827,7 +827,7 @@ let check_symbolic_values (meta : Meta.meta) (ctx : eval_ctx) : unit =
       sanity_check __FILE__ __LINE__ (info.env_count <= 1) meta;
     (* A duplicated symbolic value is necessarily primitively copyable *)
```
sonmarcho, line 828:
```suggestion
    (* A duplicated symbolic value is necessarily copyable *)
```

sonmarcho (approved):
Ready to merge after you update the flake.nix to use the main branch of Charon

Issue #84, opened 2024-03-08
Update the ways CPS is used in the compiler


sonmarcho:
The Aeneas compiler is implemented in CPS because we need to progressively build the execution trace (which is in essence a tree with holes). The way the CPS style is currently implemented makes it very hard to get debugging traces when a failure occurs.

PR #120, opened 2024-04-11
Update the interpreter so that it is not written in CPS style

EschericHya:


sonmarcho:
This PR addresses https://github.com/AeneasVerif/aeneas/issues/84

PR #121, opened 2024-04-11
Reformat the code

sonmarcho:


Nadrieril:
We should probably check that formatting is ok in CI

Nadrieril:
Oh I just saw the issue :D

PR #126, opened 2024-04-12
lean: scalars form a preorder

RaitoBezarius:
Via the canonical injection, we can easily define an induced preorder on scalars and inherit all nice properties.

It's useful to reason on specific scalar preorders w.r.t. Ordering, see the binary search tree verification example.


sonmarcho:
Looks good, thanks for this!

sonmarcho (approved):

PR #127, opened 2024-04-15
Fix CI

Nadrieril:
`flake.nix` tried to run `make tests`, which doesn't exist (it's `make test` since https://github.com/AeneasVerif/aeneas/pull/93). This resulted in "nothing to be done for `tests`", which didn't error because there is a directory called `tests` x))

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,10 @@ clean:
 
 # Test the project by translating test files to F*
 .PHONY: test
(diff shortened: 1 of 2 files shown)


sonmarcho:
It is extremely embarrassing to set up a CI and realise that it actually doesn't fail when it ought to... Thanks a lot for this: that's a nice catch!

Issue #128, opened 2024-04-15
Update the CI to check there is no diff in the generated files


sonmarcho:
The CI should already do that, but it seems it doesn't work.
For instance, the CI for https://github.com/AeneasVerif/aeneas/pull/125 should have failed.

Nadrieril:
What's the easiest way to detect which files are generated? I'd like to run something like `make clean-generated-files` in CI to be sure we regenerate everything well.

Nadrieril:
Oh, I can grep for `THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS` :D

This reveals that a number of files in the repo are not in fact regenerated, in particular a bunch of `hol4` ones.

PR #132, opened 2024-04-16
Ensure we regenerate files properly in CI

Nadrieril:
Fixes https://github.com/AeneasVerif/aeneas/issues/128

A number of files were not regenerated. Removing them caused CI to fail, so instead I marked them as not automatically-generated. Let me know which you want to keep and which we don't need.


Makefile
```diff
@@ -100,6 +100,11 @@ test-all: test-no_nested_borrows test-paper \
 	test-loops \
 	test-arrays test-traits test-bitwise test-demo
 
+.PHONY: clean-generated
+clean-generated:
+	# We can't put this line in `tests/Makefile` otherwise it will detect itself :D
```
sonmarcho, line 105:
```suggestion
	# We can't put this line in `tests/Makefile` otherwise it will detect itself
```

sonmarcho: (yes, I'm boring)

sonmarcho (changes requested):
I deactivated the regeneration of the HOL4 files. Could you revert the changes for them and simply ignore the files in the `tests/hol4` folder?

sonmarcho (approved):

PR #135, opened 2024-04-16
compiler: add `core::option::Option::{take, is_none}` and `core::mem::swap` support

RaitoBezarius:
See each commit message for each details. 

I "tried" to fix my problem in https://github.com/AeneasVerif/aeneas/issues/131 to learn more about how the compiler works, step by step.

TODO:

- [x] Add things for HOL4/Coq/etc.
- [x] Test this?
- [x] Add tests or regenerate them
- [x] Do I need to do more?

Fixes #131.
Fixes #137.
Fixes #138.

PR #136, opened 2024-04-16
compiler: map `core::option::Option::is_none` to `Option.isNone`

### LINKED ISSUE #138: `Option::is_none` is unsupported
Same as #131, but for `Option::is_none`, discovered after #135, fixed by #136.

RaitoBezarius:
This adds support for `Option.is_none` in Rust code which needs to be mapped to its backend equivalent.

Fixes #138.

### DIFF
--- a/compiler/ExtractBuiltin.ml
+++ b/compiler/ExtractBuiltin.ml
@@ -290,6 +290,9 @@ let builtin_funs () : (pattern * bool list option * builtin_fun_info) list =
     mk_fun "core::slice::{[@T]}::len"
       (Some (backend_choice "slice::len" "Slice::len"))
       None;
+    mk_fun "core::option::{core::option::Option<@T>}::is_none"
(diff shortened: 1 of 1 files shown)


compiler/ExtractBuiltin.ml
```diff
@@ -290,6 +290,9 @@ let builtin_funs () : (pattern * bool list option * builtin_fun_info) list =
     mk_fun "core::slice::{[@T]}::len"
       (Some (backend_choice "slice::len" "Slice::len"))
       None;
+    mk_fun "core::option::{core::option::Option<@T>}::is_none"
+      (Some (backend_choice "option::is_none" "Option::isNone"))
+      (Some [ false ]);
```
RaitoBezarius, line 295: FUTUREWORK: This is a hack to avoid having an alias function to take the type explicitly, on the long run, the type should be implicit and we won't have to use the filter field as a hack.

RaitoBezarius:
Closed in favor of #135.

PR #146, opened 2024-04-17
fix(backends/lean): add a significant amount of keywords

### LINKED ISSUE #139: Usage of `as` in Lean can cause syntactical errors
Consider:

```lean
/- [avl_verification::{avl_verification::AVLTreeSet<T>#2}::insert_phase1]:
   Source: 'src/main.rs', lines 199:4-199:49 -/
def AVLTreeSet.insert_phase1
  (T : Type) (OrdInst : Ord T) (self : AVLTreeSet T) (value : T) :
  Result (Bool × (AVLTreeSet T))
  :=
  do
  let (b, as) ← AVLTreeSet.insert_phase1_loop T OrdInst value self.root
  Result.ok (b, { root := as })
```

`as` seems to be a keyword or something causing:
```
▶ 280:10-280:13: error:
unexpected token 'as'; expected term
```

Renaming `as` to `u` fixes the issue.

RaitoBezarius:
This started off from the fact that `as` is a reserved keyword and cannot be used as a variable name.
And this ended up adding many keywords.

Fixes #139.

cc @sonmarcho


sonmarcho:
Can you also add "from" as a keyword?

sonmarcho:
> If we're in the process of fixing Lean keyword issues, should we complete the list using, e.g., https://github.com/leanprover/lean4/blob/master/doc/latex/lstlean.tex#L12 ?

Yes, that's a good idea. I was not aware of this list.

RaitoBezarius:
I'm doing it.

RaitoBezarius:
Done @sonmarcho @R1kM PTAL.

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "attribute";
+          "axiom";
+          "axioms";
+          "begin";
+          "break";
           "by";
+          "calc";
+          "cases";
+          "catch";
+          "check";
```
sonmarcho, line 940: I think we can remove this one

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
           "class";
+          "clear";
```
sonmarcho, line 942: This is a tactic? I think we can remove it

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
         [
+          "Pi";
+          "Prop";
+          "Sort";
+          "Type";
+          "abbrev";
+          "add_key_equivalence";
+          "alias";
+          "allGoals";
+          "apply";
```
sonmarcho, line 926: remove allGoals and apply: those are tactics

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "as";
+          "assume";
```
sonmarcho, line 928: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "catch";
+          "check";
           "class";
+          "clear";
+          "const";
+          "constant";
+          "constants";
+          "constructor";
+          "continue";
+          "contradiction";
```
sonmarcho, line 948: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "declare_syntax_cat";
+          "declare_trace";
           "decreasing_by";
           "def";
+          "definition";
           "deriving";
+          "dif";
```
sonmarcho, line 955: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
           "do";
+          "done";
           "else";
           "end";
+          "eval";
+          "exact";
```
sonmarcho, line 961: remove those two

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "example";
+          "exists";
+          "export";
+          "extends";
           "for";
+          "forall";
+          "from";
+          "fun";
+          "generalize";
```
sonmarcho, line 970: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
           "have";
+          "help";
```
sonmarcho, line 972: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "precedence";
+          "prefix";
+          "prelude";
+          "print";
+          "private";
+          "protected";
+          "raw";
+          "record";
+          "reduce";
+          "refine";
```
sonmarcho, line 1017: remove

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "simp_all";
           "structure";
+          "suffices";
           "syntax";
           "termination_by";
           "then";
-          "Type";
+          "theorem";
+          "theory";
+          "this";
```
sonmarcho, line 1040: remove

sonmarcho (changes requested):
There are several keywords which are tactics, or which require to be prefixed with `#` (I think), so it should be ok to remove them.

compiler/ExtractBase.ml
```diff
@@ -915,38 +915,140 @@ let keywords () =
...
+          "raw";
+          "record";
+          "reduce";
+          "refine";
+          "register_simp_ext";
+          "renaming";
+          "repeat";
+          "replacing";
+          "reserve";
+          "return";
```
sonmarcho, line 1023: Remove this one: I'm actually using it

RaitoBezarius:
> There are several keywords which are tactics, or which require to be prefixed with `#` (I think), so it should be ok to remove them.

I removed all the tactics I could find, PTAL.

sonmarcho:
I think we can also remove `check` and `help` because they are commands (which must be prefixed by `#`)?

RaitoBezarius:
> I think we can also remove `check` and `help` because they are commands (which must be prefixed by `#`)?

Apologies, I must have missed it, this is done now.

sonmarcho (approved):

PR #149, opened 2024-04-18
chore: explain a Nix-powered workflow

RaitoBezarius:
To avoid divergence between Charon and Aeneas, we should re-export Charon via our Flake and tell users to use this as a source of truth.

Here's an appendix on how I do refresh of my files, which can serve as inspiration for a quick start workflow.

cc @Nadrieril @sonmarcho

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,25 @@ and tactics specialized for monadic programs (see
 
 A tutorial for the Lean backend is available [here](./tests/lean/Tutorial.lean).
 
+## Quick start for Nix users
+
+Assuming Nix is installed, with a support for Flakes (`*`):
+
+```console
+$ # Run Charon with the exact same version required by Aeneas
+$ nix run github:aeneasverif/aeneas#charon -L
+$ nix run github:aeneasverif/aeneas -L -- -backend your_preferred_backend your_llbc.file
+```
+
+To regenerate the extraction, just run step 2 and step 3 again.
+
+`(*)` : Flakes are not necessary, here is an example of how to do similar steps without it:
+
+```console
+$ nix-shell '<aeneas>' -A packages.x86_64-linux.charon --run "charon" -I aeneas=https://github.com/AeneasVerif/aeneas/archive/main.tar.gz
+$ nix-shell '<aeneas>' -A packages.x86_64-linux.default --run "aeneas --backend your_preferred_backend your_llbc.file" -I aeneas=https://github.com/AeneasVerif/aeneas/archive/main.tar.gz
+```
+
 ## Formalization
 
 The translation has been formalized and published at ICFP2022: [Aeneas: Rust
--- a/flake.nix
+++ b/flake.nix
@@ -217,6 +217,7 @@
       {
         packages = {
           inherit aeneas;
+          inherit (charon.packages.${system}) charon-ml charon;
           default = aeneas;
         };
         devShells.default = pkgs.mkShell {


Nadrieril:
I'm working onremoving the need for `rustup`

sonmarcho:
What was the issue exactly?

RaitoBezarius:
> What was the issue exactly?

It's easy to delude myself (anyone-self?) to use Charon and Aeneas separately and have difficult to debug issues because Charon and Aeneas are non-synchronized.

This workflow explains how to avoid this _by design_ by reusing the Charon that Aeneas is depending on.

README.md
```diff
@@ -120,6 +120,29 @@ and tactics specialized for monadic programs (see
...
+$ nix-shell -p rustup
+$ nix run github:aeneasverif/aeneas#charon -L
+$ # The default package for Aeneas is Aeneas.
+$ nix run github:aeneasverif/aeneas -L -- -backend your_preferred_backend your_llbc.file
+```
+
+To regenerate the extraction, just run step 2 and step 3 again.
+
+It is critical not to use the Charon repository and rely on the locking
+performed in this repository to avoid incompatible LLBC files.
```
sonmarcho, line 137: Explain that sometimes it may happen that Charon and Aeneas are desynchronized: for instance some changes to the LLBC in the side of Charon haven't been merged to Aeneas yet. We try to avoid this as much as possible, but in you're using Nix, there is a simple way of avoiding the problem, etc.

Nadrieril:
Btw, rustup should no longer be needed

README.md
Nadrieril: See also https://github.com/AeneasVerif/charon/issues/35, which I've got a plan for if I ever find the time

Nadrieril:
Actually, why not add `charon` to the dev shell? You'd have to find a way to male `charon-ml` available to opam. The idea being that someone who only wants to dev on aeneas doesn't need to figure out how to build charon. Those who dev on both an have a special override

RaitoBezarius:
> Actually, why not add `charon` to the dev shell? You'd have to find a way to male `charon-ml` available to opam. The idea being that someone who only wants to dev on aeneas doesn't need to figure out how to build charon. Those who dev on both an have a special override

Seems orthogonal to me? People will `nix run ...`, not `nix develop ...` in general.
Or they would do what I did in https://github.com/RaitoBezarius/avl-verification/blob/main/default.nix#L9-L15 for longer term setups IMHO.

Nadrieril:
Oh you're right, I was thinking of aeneas devs but this PR is for aeneas users. Nevermind

README.md
sonmarcho: Yes, I agree this is pertinent. I just want the comments to be modified to take into account my remarks above, and then I approve the PR. :)

RaitoBezarius: Addressed!

README.md
```diff
@@ -120,6 +120,30 @@ and tactics specialized for monadic programs (see
 
 A tutorial for the Lean backend is available [here](./tests/lean/Tutorial.lean).
 
+## Quick start for Nix users
+
+Assuming Nix is installed, with a support for Flakes (`*`):
+
+```console
+$ nix run github:aeneasverif/aeneas#charon -L
```
Nadrieril, line 128: Could you add a comment like "run charon with the exact version required by aeneas" above?

Nadrieril (approved):

RaitoBezarius:
Wait, don't merge. This won't work without rustup, I think.

README.md
RaitoBezarius: Yep, can do that.

RaitoBezarius: But wait, running charon like this will run it with the version of aeneas, no?

Nadrieril:
Oh yeah this needs latest charon; I'll bump soon

README.md
Nadrieril: Yes but this may not be obvious, hence why I suggest adding a comment that explains what's happening

README.md
```diff
@@ -120,6 +120,30 @@ and tactics specialized for monadic programs (see
...
+$ # The default package for Aeneas is Aeneas.
```
Nadrieril, line 129: I think this comment is more confusing than not having it :p

README.md
```diff
@@ -120,6 +120,30 @@ and tactics specialized for monadic programs (see
...
+$ nix run github:aeneasverif/aeneas -L -- -backend your_preferred_backend your_llbc.file
+```
+
+To regenerate the extraction, just run step 2 and step 3 again.
+
+Sometimes, Charon and Aeneas can become desynchronized, changes of the LLBC on
+Charon side will not have been taken into account on Aeneas side yet. We try to
+avoid this as much but if you are using Nix, this is the simple way to avoid
+that problem. https://github.com/AeneasVerif/charon/issues/35 is related to that matter.
```
Nadrieril, line 138: We now have version numbers. This paragraph still makes sense but you can mention the new behavior.

Nadrieril: Actually, this won't happen if the users uses the charon exported by aeneas, so you can remove the paragraph.

README.md
RaitoBezarius: Removed.

README.md
RaitoBezarius: Removed.

RaitoBezarius:
I need to do a test with that workflow, I recently had some issues while trying to extract stuff.

Nadrieril (approved):

sonmarcho (approved):
Looks good, thanks!

Issue #152, opened 2024-04-22
Don't run the same CI jobs twice


sonmarcho:
Same as in Charon. This would be extremely useful for the Lean tests which don't go through nix (they are expensive and are really run twice).

PR #155, opened 2024-04-22
ci: avoid running duplicate jobs

Nadrieril:
Does the same thing as https://github.com/AeneasVerif/charon/pull/113

Fixes https://github.com/AeneasVerif/aeneas/issues/152


sonmarcho (approved):
That's great, thanks!

PR #157, opened 2024-04-24
chore(*): add pre-commit hooks for formatting

RaitoBezarius:
This adds pre-commit hooks for formatting, because I usually just trust whatever people prefer locally :).
We could add, e.g. rustfmt, but I prefer @sonmarcho that you tell me what you prefer regarding this. Pre-commit hooks can be pesky and annoying.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,11 +9,12 @@
     nixpkgs.follows = "charon/nixpkgs";
     hacl-nix.url = "github:hacl-star/hacl-nix";
     flake-compat.url = "github:nix-community/flake-compat";
+    pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
   };
 
   # Remark: keep the list of outputs in sync with the list of inputs above
   # (see above remark)
-  outputs = { self, charon, flake-utils, nixpkgs, hacl-nix, flake-compat }:
+  outputs = { self, charon, flake-utils, nixpkgs, hacl-nix, flake-compat, pre-commit-hooks }:
     flake-utils.lib.eachSystem [ "x86_64-linux" ] (system:
       let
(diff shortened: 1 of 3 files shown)


Nadrieril (approved):
We recently added CI checks for formatting, this makes a lot of sense. Could you do the same (with rustfmt also) on Charon's side?

RaitoBezarius:
> We recently added CI checks for formatting, this makes a lot of sense. Could you do the same (with rustfmt also) on Charon's side?

Yes, of course!

sonmarcho:
In practice what does this do? It only uses pre commit hooks if you're using nix, right? If that's the case and you consider it's idiomatic when using nix, I'm fine.

flake.nix
```diff
@@ -183,9 +189,26 @@
...
-                  aeneas-verify-fstar
-                  aeneas-verify-coq
-                  aeneas-verify-hol4
-                  aeneas-check-tidiness; };
+            aeneas-verify-fstar
+            aeneas-verify-coq
+            aeneas-verify-hol4
+            aeneas-check-tidiness;
+
+          pre-commit-check = pre-commit-hooks.lib.${system}.run {
```
Nadrieril, line 197: I'm not a huge fan of pre-commit hooks because I rebase a lot... I does make sense though, we can try it for a bit

sonmarcho: Can you add a comment here to explain what this does (this formats the code at commit time, etc.)?

Nadrieril:
Now that my editor formats everything automatically I'm not longer keen on having this

sonmarcho:
This PR seems dead, so I'm closing it. Feel free to reopen it if you really want this feature.

PR #158, opened 2024-04-24
chore(ci): move Lean CI under Nix

RaitoBezarius:
As it often happens, the Lean CI under Ubuntu is broken: https://github.com/AeneasVerif/aeneas/actions/runs/8814059410/job/24193132680?pr=135 and blocking PRs.

Lean doesn't work nicely under the Nix sandbox, but in a CI context, we can impurely run scripts and use Nix to get our dependencies, e.g. curl or elan in this case.

It is still more reliable than letting Ubuntu or GitHub Actions figure out their signing for their APT repositories apparently.


RaitoBezarius:
@Nadrieril can you press the button :<

.github/workflows/ci.yml
```diff
@@ -28,30 +28,24 @@ jobs:
...
-      # Checkout the repo and download it to the runner
-      - name: Checkout
-        uses: actions/checkout@v4
-      # Verify - note that we need to update the environment with `source` so
-      # that the lake binary is in the path.
-      - run: source ~/.profile && cd tests/lean && make
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v26
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
```
Nadrieril, line 49: Is there an easy way that you can reuse the nixpkgs we use for the other tests?

RaitoBezarius: What I could do is simply have a devShells with `elan` from our flake and go from there.

Nadrieril:
I think you broke the yaml with a syntax error so github doesn't see any jobs to run

.github/workflows/ci.yml
RaitoBezarius: (But I'm mostly interested into validating if this idea can work so, using our latest nixpkgs should not matter too much.)

RaitoBezarius:
> I think you broke the yaml with a syntax error so github doesn't see any jobs to run

And this is why I would like to write .nix instead of yml for GitHub actions :).

.github/workflows/ci.yml
Nadrieril: Apparently you need dun too, so devshell seems like the way to go

Nadrieril:
I think you want: `nix develop --command bash -c "cd tests/lean && make"`

.github/workflows/ci.yml
```diff
@@ -28,30 +28,23 @@ jobs:
...
-      # Checkout the repo and download it to the runner
-      - name: Checkout
-        uses: actions/checkout@v4
-      # Verify - note that we need to update the environment with `source` so
-      # that the lake binary is in the path.
-      - run: source ~/.profile && cd tests/lean && make
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v26
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
```
Nadrieril, line 49: this isn't needed anymore

RaitoBezarius: That is true

RaitoBezarius: Done

.github/workflows/ci.yml
```diff
@@ -28,30 +28,21 @@ jobs:
...
       - run: nix build -L .#checks.x86_64-linux.aeneas-verify-fstar
       - run: nix build -L .#checks.x86_64-linux.aeneas-verify-coq
       - run: nix build -L .#checks.x86_64-linux.aeneas-verify-hol4
-      # Lean doesn't work with Nix
+      # Lean cannot run its tests in the sandbox because `elan` will download things
       #- run: nix build -L .#checks.x86_64-linux.aeneas-verify-lean
 
-  lean: # Lean isn't supported by Nix, so we put it in a different job
+  lean:
     runs-on: [ubuntu-latest]
```
Nadrieril, line 42: why not run it on our self-hosted machine?

RaitoBezarius: you lost parallelism

Nadrieril: fair

Nadrieril (approved):

sonmarcho:
Could you provide more context as to why the CI broke and why it is more stable to do this with Nix? Because my understanding is that the download part (which is the unstable part) is still done *outside of Nix* in your PR?

sonmarcho:
> > Could you provide more context as to why the CI broke and why it is more stable to do this with Nix? Because my understanding is that the download part (which is the unstable part) is still done _outside of Nix_ in your PR?
> 
> [eb519e0](https://github.com/AeneasVerif/aeneas/commit/eb519e0e56d30763db6ca6667a9f4095061252a4) it's written in the commit message.
> 
> What was done was to use `apt update` with Ubuntu, the image of Ubuntu is outdated and broken and cannot update. Now, what we do is to `nix-shell -p elan` with Nix, we don't need a good image or something, we just need a working Nix installation.

I think the important missing information is that `apt update` is less stable than `nix-shell -p elan`, correct?

sonmarcho:
I rechecked the changes, it is indeed a lot cleaner than before. Merging as a consequence.

PR #164, opened 2024-04-30
feat(backend/lean): Raw Lean literals can be parsed into scalars automatically

RaitoBezarius:
We can use coercion to go to integers, build a scalar out of an int via
unification and let the user clear up bounds or let Lean decide them.

Untested yet.

### DIFF
--- a/backends/lean/Base/Primitives/Scalar.lean
+++ b/backends/lean/Base/Primitives/Scalar.lean
@@ -332,6 +332,11 @@ instance [Decide (Scalar.cMin ty ≤ v ∧ v ≤ Scalar.cMax ty)] : InBounds ty 
 @[simp] abbrev Scalar.check_bounds (ty : ScalarTy) (x : Int) : Bool :=
   (Scalar.cMin ty ≤ x || Scalar.min ty ≤ x) ∧ (x ≤ Scalar.cMax ty || x ≤ Scalar.max ty)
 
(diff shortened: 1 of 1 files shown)


sonmarcho:
This is a good remark, especially as since recently we can solve the bound obligations by using the trait mechanism. I'm still a bit reluctant to making the machine integer kind completely implicit for constants is that I intend to make a modification soon to make as many type as possible implicit in function signatures. For the design I have in mind we will need to make sure all the inputs to the functions are unambiguously typed, which is not the case anymore if we do this modification.

RaitoBezarius:
Hmm, from a compiler perspective, I suppose that everything can always manually be coerced into the right type, no? So there would not be ambiguity on that side.

From an end-user perspective, yes, they could manually coerce it but also this can automagically coerce it.

No strong opinion, I just came to find this handy.

sonmarcho:
I need to think about the compiler side, but it's indeed a good idea to provide this to the user side.
I note you only added and instance of `OfNat`. Does the `OfInt` typeclass also exist in Lean?

RaitoBezarius:
> I need to think about the compiler side, but it's indeed a good idea to provide this to the user side. I note you only added and instance of `OfNat`. Does the `OfInt` typeclass also exist in Lean?

In the Prelude, I'm only aware of `OfNat`: https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#OfNat

sonmarcho (approved):

PR #166, opened 2024-05-01
Propagate handling of (unsupported) mutually recursive trait declarations

R1kM:
Companion PR of https://github.com/AeneasVerif/charon/pull/160

This PR propagates the extension of mutually recursive traits in Charon, now raising an error in Aeneas during translation as this feature is currently unsupported.

Marking this as draft for now for two reasons:
* The error reporting should likely be enhanced with meta information, however, I am not sure how to retrieve it in the current context (hence the temporary use of `craise_meta_opt`)
* As I extended the Charon traits tests with mutually recursive trait declarations, this leads to failures in Aeneas when running `make tests`. Maybe a better way is to split the Charon test file into what is supported and unsupported in Aeneas?

### DIFF
--- a/compiler/Translate.ml
+++ b/compiler/Translate.ml
@@ -798,7 +798,10 @@ let extract_definitions (fmt : Format.formatter) (config : gen_config)
         (* Translate *)
         export_functions_group pure_funs
     | GlobalGroup id -> export_global id
-    | TraitDeclGroup id ->
+    | TraitDeclGroup (RecGroup _ids) ->
+        craise_opt_meta __FILE__ __LINE__ None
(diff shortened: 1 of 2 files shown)


Nadrieril:
The tests that are meant for Charon should be put in `charon/tests/ui`. The tests in `tests/` and `tests-polonius/` are meant to be soon moved to aeneas.

R1kM:
Marking as ready as the companion PR has been merged in Charon.
@sonmarcho  Let's improve the error message in a follow-up PR.

sonmarcho:
I created an issue: https://github.com/AeneasVerif/aeneas/issues/176

PR #183, opened 2024-05-15
feat(backends/lean): make `max`-related coercions nicer

RaitoBezarius:
Situations where you have `coe (max a b) = max (coe a) (coe b)` are often stuck during verification because of the lack of this theorem.

With this theorem, `push_cast` works as intended and normalizes even further.

### DIFF
--- a/backends/lean/Base/Primitives/Scalar.lean
+++ b/backends/lean/Base/Primitives/Scalar.lean
@@ -1419,6 +1419,18 @@ instance (ty: ScalarTy) : LinearOrder (Scalar ty) where
     right; exact (Scalar.le_equiv _ _).2 H
   decidableLE := ScalarDecidableLE ty
 
+-- Coercion theorems
+-- This is helpful whenever you want to "push" casts to the innermost nodes
+-- and make the cast normalization happen more magically.
+
+@[simp, norm_cast]
+theorem coe_max {ty: ScalarTy} (a b: Scalar ty): ↑(Max.max a b) = (Max.max (↑a) (↑b): ℤ) := by
+  -- TODO: there should be a shorter way to prove this.
+  rw [max_def, max_def]
+  split_ifs <;> simp_all
+  refine' absurd _ (lt_irrefl a)
+  exact lt_of_le_of_lt (by assumption) ((Scalar.lt_equiv _ _).2 (by assumption))
+
 -- Leading zeros
 def core.num.Usize.leading_zeros (x : Usize) : U32 := sorry
 def core.num.U8.leading_zeros (x : U8) : U32 := sorry


backends/lean/Base/Primitives/Scalar.lean
```diff
@@ -1419,6 +1419,19 @@ instance (ty: ScalarTy) : LinearOrder (Scalar ty) where
     right; exact (Scalar.le_equiv _ _).2 H
   decidableLE := ScalarDecidableLE ty
 
+-- Coercion theorems
+-- This is helpful whenever you want to "push" casts to the innermost nodes
+-- and make the cast normalization happen more magically.
+
+@[simp, norm_cast]
+theorem coe_max {ty: ScalarTy} (a b: Scalar ty): Max.max a b = (Max.max a b: ℤ) := by
+  -- TODO: redo me better because this is a bit weird.
```
sonmarcho, line 1428: Redo what? The proof?

backends/lean/Base/Primitives/Scalar.lean
```diff
@@ -1419,6 +1419,19 @@ instance (ty: ScalarTy) : LinearOrder (Scalar ty) where
     right; exact (Scalar.le_equiv _ _).2 H
   decidableLE := ScalarDecidableLE ty
 
+-- Coercion theorems
+-- This is helpful whenever you want to "push" casts to the innermost nodes
+-- and make the cast normalization happen more magically.
+
+@[simp, norm_cast]
+theorem coe_max {ty: ScalarTy} (a b: Scalar ty): Max.max a b = (Max.max a b: ℤ) := by
```
sonmarcho, line 1427: What does this theorem do?
On the left a and b are scalars, and on the right they are coerced to integers first, correct?
If yes, it would make more sense to put the coercion on a and b (rather than `Max.max a b`).

sonmarcho:
You confirm that you tested this theorem and it allows you to do the proofs in practice?

RaitoBezarius:
Yep, it does help me to progress in proof.

backends/lean/Base/Primitives/Scalar.lean
RaitoBezarius: Yeah, I was not very happy with the body of the proof, I need to take a look again.

backends/lean/Base/Primitives/Scalar.lean
RaitoBezarius: `Max.max (coe a) (coe b) = coe (Max.max a b)`, this is what this theorem does.

RaitoBezarius: OK, I will put explicitly the coercion in the arguments as well.

RaitoBezarius: @sonmarcho done.

backends/lean/Base/Primitives/Scalar.lean
RaitoBezarius: Maybe we can merge this for the time being and revisit that proof later or I can remove the TODO, let me know what you prefer

sonmarcho (approved):

PR #184, opened 2024-05-15
feat: add small pieces of max theory

RaitoBezarius:
`0#ty` is neutral for `max` for unsigned integers.


backends/lean/Base/Primitives/Scalar.lean
```diff
@@ -1419,6 +1423,21 @@ instance (ty: ScalarTy) : LinearOrder (Scalar ty) where
...
+
+-- Max theory
+-- TODO: do the min theory later on.
+
+theorem Scalar.zero_le_unsigned {ty} (s: ¬ ty.isSigned) (x: Scalar ty): Scalar.ofInt 0 ≤ x := by
+  apply (Scalar.le_equiv _ _).2
+  convert x.hmin
+  cases ty <;> simp [ScalarTy.isSigned] at s <;> simp [Scalar.min]
+
+theorem Scalar.max_unsigned_left_zero_eq {ty} (s: ¬ ty.isSigned) (x: Scalar ty):
```
sonmarcho, line 1435: Do you confirm that simplification with this theorem works? In particular, is the `¬ ty.isSigned` automatically discharged when you expect it to be discharged?

RaitoBezarius: No, I think we determined it doesn't, it must be provided manually.

sonmarcho: I think we had figured out a potential solution, by using typeclasses?

RaitoBezarius: Yep, we need to use the `Fact` typeclass to have that predicate around.

RaitoBezarius: @sonmarcho I pushed a version with the fact typeclass, should we do the fact instances in a upcoming PR? I am not exactly sure how we want to do them, e.g. just copy/paste `instance : Fact (not ty.isSigned) where out := ...` for all `ty` or just do a bit of metaprog to generate all those instances with a similar tactic?

sonmarcho: I'd rather do them in this PR: I want to get those simp lemmas working before merging anything. Having one instance per unsigned integer type would work.

RaitoBezarius: OK, I can do that

RaitoBezarius: This is done, @sonmarcho!

sonmarcho: Excellent, thanks! :)

sonmarcho:
There are conflicts with `main` now it seems

RaitoBezarius:
I will rebase.

Issue #182, opened 2024-05-15
Make the test suite standalone


Nadrieril:
Today we rely on tests from `charon` for our own tests. It would be better to have the rust test files here and adapt the scripts to call `charon` as needed.

PR #194, opened 2024-05-23
Support field projectors for recursive structs in Lean backend

### LINKED ISSUE #185: Generate projectors for recursive structs during Lean extraction
As described in #178, recursive Rust structs are extracted to Lean inductives due to Lean not supporting recursion in structures.
However, to retrieve similar functionalities, it would be useful to always extract projectors in this specific case, which is already done for the Coq backend.

R1kM:
Lean structures cannot be recursive. For this reason, recursive Rust structs are extracted to inductives with one constructor.

Similarly to what is already done for Coq, this PR generates projectors which allows to retrieve the expected behavior of structs once extracted to Lean. The translation function is similar to the one for Coq, however, there are some minor differences throughout the function, which is why I duplicated the code instead of making `extract_type_decl_lean_record_field_projectors` backend-generic.

This PR also removes the rewriting of Lean field projectors for recursive structs into let-bindings, to instead use the newly generated projectors.

I tested this feature with the following Rust code, which should be added to the testsuite once the refactoring of Aeneas tests (#182) is complete.

```rust
struct AVLNode<T> {
    value: T,
    left: AVLTree<T>,
    right: AVLTree<T>,
}

type AVLTree<T> = Option<Box<AVLNode<T>>>;


fn get_val<T>(x: AVLNode<T>) -> T  { x.value }

fn get_left<T>(x: AVLNode<T>) -> AVLTree<T> { x.left }
```

Fixes #185


sonmarcho:
This looks good. I just want to investigate the problem of generating simplification lemmas (see https://github.com/AeneasVerif/aeneas/issues/185) before merging this.

R1kM:
It seems that for structures, Lean 4 automatically adds the attribute `reducible` to projectors, which allows to retrieve the behavior you wanted (i.e., simplify even with `simp only []`).
Maybe we could just add this attribute to generated projectors?

sonmarcho:
Following some experimentations, it seems we need to add `@[reducible]` and we also need to register explicitly register the projection functions for `simp` to reduce them. The function which does so is `Lean.addProjectionFnInfo`.

sonmarcho:
There is an interesting discussion here as to why recursive structures are not supported:
https://github.com/leanprover/lean4/issues/4182

sonmarcho:
After some discussions: we will mark the projectors as `reducible` and `simp` for the time being.

R1kM:
> After some discussions: we will mark the projectors as `reducible` and `simp` for the time being.

This is done in [e76ef95](https://github.com/AeneasVerif/aeneas/pull/194/commits/e76ef95ddda11d4220e1f0fd863b0df568de95bc)

Issue #196, opened 2024-05-23
Test different the approaches to statefulness


Nadrieril:
In a coming PR, I will remove a duplicated test that was added to test `-backward-no-state-update`. We should add a test back to make sure we exercise all the codegen options.

sonmarcho:
This is going to be subsumed by the support for separation logic.

PR #197, opened 2024-05-23
Standalone test harness

Nadrieril:
This fairly long PR gives a clean testing story to aeneas. With this PR, we no longer depend on the charon test suite. Instead, any file in `tests/src` is automatically processed when we do `make test`. For now, the options for each file are hardcoded in the test runner. Eventually we'll be able to set file-specific options at the top of the each file like in the charon test suite.

In the process, I removed the duplicated betree test (to make my life simpler). I opened https://github.com/AeneasVerif/aeneas/issues/196 so we can add a proper test for whatever we wanted to test there.


Nadrieril:
This mostly solves #197 but I want to add `//@` comment parsing before I consider #197 complete.

tests/test_runner/run_test.ml
```diff
@@ -0,0 +1,233 @@
...
+          path = test_path;
+          name = test_name;
+          kind = charon_kind;
+          extra_charon_options;
+        }
+      in
+      (* Generate the llbc file *)
+      run_charon runner_env charon_case;
+
+      (* FIXME: remove this special case *)
```
sonmarcho, line 213: The issue is only that we need to rename the folders, correct?

tests/test_runner/run_test.ml
```diff
@@ -0,0 +1,233 @@
...
+  | "lean", "demo" -> "Demo"
+  | "lean", _ -> "."
+  | _, ("arrays" | "demo" | "hashmap" | "traits") -> test_name
+  | _, "betree_main" -> "betree"
+  | _, "hashmap_main" -> "hashmap_on_disk"
+  | "hol4", _ -> "misc-" ^ test_name
+  | ( _,
+      ( "bitwise" | "constants" | "external" | "loops" | "no_nested_borrows"
+      | "paper" | "polonius_list" ) ) ->
+      "misc"
```
sonmarcho, line 185: I'm ok with getting rid of the "misc" folder and having one folder per test. It would be cumbersome to maintain on the Rust side, but as the files are generated it's ok. Also, I think it's actually cleaner: it made sense to have this "misc" folder when I generated only one file per test, but there are today tests which generate several files (for instance the "loops" test).

tests/test_runner/run_test.ml
```diff
@@ -0,0 +1,233 @@
...
+
+(* The subdirectory in which to store the outputs. *)
+(* This reproduces the file layout that was set by the old Makefile. FIXME: cleanup *)
+let test_subdir backend test_name =
+  match (backend, test_name) with
+  | "lean", "demo" -> "Demo"
+  | "lean", _ -> "."
+  | _, ("arrays" | "demo" | "hashmap" | "traits") -> test_name
+  | _, "betree_main" -> "betree"
+  | _, "hashmap_main" -> "hashmap_on_disk"
```
sonmarcho, line 180: In the future I would like to merge the hashmap_on_disk test with the hashmap test. It should be doable once we have attributes to mark definitions as opaque, and will allow us to only have one file for the hashmap.

sonmarcho (approved):
This was badly needed. There are still some changes to do but we can keep them for a subsequent PR. Thanks!

Makefile
```diff
@@ -100,207 +99,88 @@ check-charon:
...
+
+# =============================================================================
+# The tests.
+# =============================================================================
+
+# Test the project by translating test files to various backends.
+.PHONY: test
+test: build-dev test-all betree-tests
+
+# This runs the rust tests of the betree crate.
```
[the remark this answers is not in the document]
Nadrieril: That's not what's happening here: there's no special rule for building the llbc for betree (the test runner does the appropriate magic). This here is to run `cargo test` in the crate.

tests/test_runner/run_test.ml
Nadrieril: Noted in #208

tests/test_runner/run_test.ml
Nadrieril: Noted in #208

PR #204, opened 2024-05-24
Add magic comments to test runner

### LINKED ISSUE #182: Make the test suite standalone

Nadrieril:
It is now possible to specify charon and aeneas options directly in the rust source file, like in the charon test suite.

Fixes #182

### DIFF
--- a/tests/test_runner/run_test.ml
+++ b/tests/test_runner/run_test.ml
@@ -5,22 +5,38 @@ type runner_env = {
   llbc_dir : string;
 }
 
-(* The data for a specific test to run aeneas on *)
-type aeneas_test_case = {
-  name : string;
-  backend : string;
-  subdir : string;
-  extra_aeneas_options : string list;
-}
+module Backend = struct
+  type t = Coq | Lean | FStar | HOL4 [@@deriving ord, sexp]
+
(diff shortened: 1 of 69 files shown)


tests/test_runner/run_test.ml
```diff
@@ -42,21 +58,175 @@ let run_command args =
...
+  | "lean", _ -> "."
+  | _, ("arrays" | "demo" | "hashmap" | "traits") -> test_name
+  | _, "betree" -> "betree"
+  | _, "hashmap_main" -> "hashmap_on_disk"
+  | "hol4", _ -> "misc-" ^ test_name
+  | ( _,
+      ( "bitwise" | "constants" | "external" | "loops" | "no_nested_borrows"
+      | "paper" | "polonius_list" ) ) ->
+      "misc"
+  | _ -> test_name
```
sonmarcho, line 103: I think you can put those in the magic options?

sonmarcho (approved):

tests/test_runner/run_test.ml
Nadrieril: I would prefer to clean them up so we don't need them at all

PR #209, opened 2024-05-27
Support negative tests

### LINKED ISSUE #171: Add Negative Tests
Similarly to what is done in Charon, we should add negative tests to Aeneas, for instance to ensure to catch and test unsupported features.
One example is #166, where it would be good to ensure that the checks for mutually recursive traits are indeed satisfied, and that Aeneas fails as expected on the corresponding examples.

Nadrieril:
Fixes #171


tests/lean/misc/MutuallyRecursiveTraits.lean
```diff
@@ -0,0 +1,6 @@
+-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
+-- [mutually_recursive_traits]
+import Base
+open Primitives
+
+namespace mutually_recursive_traits
```
sonmarcho, line 6: I don't understand : as you use `-abort-on-failure` this file shouldn't have been generated?...

Nadrieril: And yet it was, I just checked. I didn't dig any further, we can probably open an issue for that

Issue #208, opened 2024-05-24
Cleanup remaining tests


Nadrieril:
The aim is to remove all the hardcoded test options from the test runner:
- [x] find a way to specify options to the test runner for crates (e.g. `crate/Aeneas.toml`);
- [x] rename `betree_main` to betree;
- [x] cleanup `hashmap` vs `hashmap_main`.

Issue #195, opened 2024-05-23
Add test for recursive structs projectors


R1kM:
We should add the following test, coming from #194 

```rust
struct AVLNode<T> {
    value: T,
    left: AVLTree<T>,
    right: AVLTree<T>,
}

type AVLTree<T> = Option<Box<AVLNode<T>>>;


fn get_val<T>(x: AVLNode<T>) -> T  { x.value }

fn get_left<T>(x: AVLNode<T>) -> AVLTree<T> { x.left }
```

Nadrieril:
Fixed in https://github.com/AeneasVerif/aeneas/pull/213

Issue #191, opened 2024-05-22
Add a test with matches over integers


sonmarcho:
There are no tests exerting matches over integers:
```rust
fn match_u32(x: u32) -> u32 {
    match x {
        0 => 0,
        1 => 1,
        _ => 2,
    }
}
```

Nadrieril:
Fixed in https://github.com/AeneasVerif/aeneas/pull/213

Issue #153, opened 2024-04-22
Add a backend test for infinite loops


sonmarcho:
We should add the following test:
```rust
fn foo() {
    while true {
        bar()
    }
}
```

Nadrieril:
Fixed in https://github.com/AeneasVerif/aeneas/pull/213

PR #213, opened 2024-05-27
Cleanup tests

Nadrieril:
This PR:
- Merges `hashmap` and `hashmap_main`;
- Renames `betree_main` to betree;
- Adds a way to specify options to the test runner for crates, in `crate/aeneas-test-options`;
- Adds some tests we wanted to add.

Fixes https://github.com/AeneasVerif/aeneas/issues/208, #153, #195, #191.


tests/lean/HashmapMain/Opaque.lean
```diff
@@ -1,18 +0,0 @@
...
-  
-  /- [hashmap_main::hashmap_utils::deserialize] -/
-  hashmap_utils.deserialize_fwd
-    : State → Result (State × (hashmap_hash_map_t U64))
-  
-  /- [hashmap_main::hashmap_utils::serialize] -/
-  hashmap_utils.serialize_fwd
-    : hashmap_hash_map_t U64 → State → Result (State × Unit)
-  
-end hashmap_main
```
Nadrieril, line 18: I didn't move that file over to `Hashmap` because it didn't seem to be required. Should I move it anyway?

tests/src/infinite-loop.rs
```diff
@@ -0,0 +1,10 @@
+//@ [coq,fstar] skip
+//@ [coq,fstar] subdir=misc
+//^note: this fails on the fstar and coq backends
+fn bar() {}
+
+fn foo() {
+    loop {
+        bar()
+    }
+}
```
Nadrieril, line 10: Note that the infinite loop test fails on two backends, and the match test fails on coq. Should I open issues for that?

tests/lean/HashmapMain/Opaque.lean
sonmarcho: This is indeed not necessary: remove it

tests/src/infinite-loop.rs
sonmarcho: Where does it come from originally?
If it fails it means there are missing fails, for instance for the decrease clauses. I can look into it.

tests/src/hashmap.rs
```diff
@@ -314,6 +316,31 @@ impl<T> HashMap<T> {
     }
 }
 
+// This is a module so we can tell charon to leave it opaque
+mod hashmap_utils {
```
sonmarcho, line 320: You can rename this module to "utils"

tests/src/betree/aeneas-test-options
```diff
@@ -0,0 +1,4 @@
+charon-args=--polonius --opaque=betree_utils
+aeneas-args=-backward-no-state-update -test-trans-units -state -split-files
+[coq] aeneas-args=-use-fuel
+[fstar] aeneas-args=-decreases-clauses -template-clauses
```
sonmarcho, line 4: This file is needed because the betree is a crate. But actually we can turn it into a single file.

sonmarcho (approved):

sonmarcho:
This is great! I would like to talk about the betree before you merge however (and we can decide whether we do the change now or in a later PR).

tests/src/betree/aeneas-test-options
Nadrieril: It is a crate today because it has dependencies (notably serde). Do we want to give up on that?

PR #223, opened 2024-05-29
Propagated changes to statement from Charon

EschericHya:


compiler/PrePasses.ml
```diff
@@ -100,7 +100,7 @@ let remove_useless_cf_merges (crate : crate) (f : fun_decl) : fun_decl =
   let rec can_be_moved_aux (must_end_with_exit : bool) (st : statement) : bool =
     match st.content with
     | SetDiscriminant _ | Assert _ | Call _ | Break _ | Continue _ | Switch _
-    | Loop _ ->
+    | Loop _ | Error ->
```
sonmarcho, line 103: We should return `true` for `Error`: you can group it with `Panic` and `Return` below

sonmarcho (changes requested):

compiler/PrePasses.ml
```diff
@@ -108,7 +108,7 @@ let remove_useless_cf_merges (crate : crate) (f : fun_decl) : fun_decl =
         | Aggregate (AggregatedAdt (TTuple, _, _), []) -> not must_end_with_exit
         | _ -> false)
     | FakeRead _ | Drop _ | Nop -> not must_end_with_exit
-    | Panic | Return -> true
+    | Panic | Return | Error _ -> true
```
sonmarcho, line 111: Move the `Error` case with `SetDiscriminant`, `Assert`, etc.

sonmarcho (changes requested):

Nadrieril (approved):

PR #227, opened 2024-05-30
Remove the use of `Option` from the functions synthesizing the symbolic AST

sonmarcho:
The functions that are used to synthesize the symbolic AST, which are written as CPS style, currently take as input an optional expression and return as output an optional expression. It was useful to use options before because both the symbolic execution *and* the construction of the symbolic AST were written in CPS style, and because it was useful to be able not to generate the symbolic AST in case we run the interpreter in concrete (i.e., not symbolic) mode, or if we wanted to only borrow-check the program (this has never been implemented, but it is possible and quite easy). Today, this is not useful anymore as the interpreter is not written in CPS anymore: if we don't want to synthesize the symbolic AST (which is used only for the translation) we can simply ignore the continuations (i.e., not call them with anything). This PR takes care of this.


sonmarcho:
@R1kM I'm afraid this might conflict a bit with the PR you're working on, so I'll wait for your PR to be merged

Issue #211, opened 2024-05-27
Aeneas' projectors in Lean can collide with structure functions


RaitoBezarius:
Consider:

```rust
struct AVLTree {
    height: usize
}

impl AVLTree {
    fn height(&self) -> usize { ... }
}
```

This will yield a Lean projector and a Lean function for `height`, those will collide.

cc @R1kM @sonmarcho

sonmarcho:
This is extremely annoying. I'm not sure what the best thing to do is.

RaitoBezarius:
In Lean, we could make better use of namespaces. We have the trait and the structure.

Multiple options:

- `AVLTree.Trait.height`
- `AVLTree.Projections.height`

Let the user "import" stuff and open namespaces for their usecases.

But that's my 2 cents.

PR #239, opened 2024-06-11
Support for renaming using the rename attribute in charon

EschericHya:
Fixes https://github.com/AeneasVerif/aeneas/issues/211


compiler/ExtractBase.ml
```diff
@@ -252,6 +252,14 @@ let empty_names_map : names_map =
     names_set = StringSet.empty;
   }
 
+(** Small helper to rename a llbc_name if the rename attribute has been set *)
```
sonmarcho, line 255:
```suggestion
(** Small helper to rename an llbc_name if the rename attribute has been set *)
```

compiler/ExtractBase.ml
```diff
@@ -252,6 +252,14 @@ let empty_names_map : names_map =
...
+let rename_llbc_name (item_meta : Meta.item_meta) (llbc_name : llbc_name) =
+  match item_meta.rename with
+  | Some rename ->
+      let remove_last = List.tl (List.rev llbc_name) in
```
sonmarcho, line 259:
```suggestion
      let name_prefix = List.tl (List.rev llbc_name) in
```

compiler/ExtractBase.ml
```diff
@@ -252,6 +252,14 @@ let empty_names_map : names_map =
...
+      List.rev (T.PeIdent (rename, Disambiguator.zero) :: remove_last)
```
sonmarcho, line 260:
```suggestion
      List.rev (T.PeIdent (rename, Disambiguator.zero) :: name_prefix)
```

compiler/ExtractBase.ml
```diff
@@ -1390,7 +1398,7 @@ let ctx_compute_type_name_no_suffix (span : Meta.span) (ctx : extraction_ctx)
   flatten_name (ctx_compute_simple_type_name span ctx name)
 
 (** Provided a basename, compute a type name. *)
-let ctx_compute_type_name (span : Meta.span) (ctx : extraction_ctx)
+let ctx_compute_type_decl_name (span : Meta.span) (ctx : extraction_ctx)
```
sonmarcho, line 1401: Can you add an "_aux" suffix? I now remember why this function was named like that (there is a `ctx_compute_type_decl_name` below)

compiler/ExtractBase.ml
```diff
@@ -1409,7 +1417,7 @@ let ctx_compute_type_name (span : Meta.span) (ctx : extraction_ctx)
     access then causes trouble because not all provers accept syntax like
     [x.3] where [x] is a tuple.
  *)
-let ctx_compute_field_name (span : Meta.span) (ctx : extraction_ctx)
+let ctx_compute_field_name (def : type_decl) (ctx : extraction_ctx)
     (def_name : llbc_name) (field_id : FieldId.id) (field_name : string option)
     : string =
   let field_name_s =
```
sonmarcho, line 1423: You forgot to use the rename attribute provided for fields (below you're updating the prefix of the name we generate, in case we prefix the name of the field with the name of the structure).

compiler/ExtractBase.ml
```diff
@@ -1424,8 +1432,10 @@ let ctx_compute_field_name (span : Meta.span) (ctx : extraction_ctx)
       "_" ^ field_name_s
     else field_name_s
   else
+    let name = rename_llbc_name def.item_meta def_name in
```
sonmarcho, line 1435:
```suggestion
    let def_name = rename_llbc_name def.item_meta def_name in
```

compiler/ExtractBase.ml
```diff
@@ -1424,8 +1432,10 @@ let ctx_compute_field_name (span : Meta.span) (ctx : extraction_ctx)
...
     let def_name =
-      ctx_compute_type_name_no_suffix span ctx def_name ^ "_" ^ field_name_s
+      ctx_compute_type_name_no_suffix def.item_meta.span ctx name
```
sonmarcho, line 1437:
```suggestion
      ctx_compute_type_name_no_suffix def.item_meta.span ctx def_name
```

compiler/ExtractBase.ml
```diff
@@ -1535,7 +1546,10 @@ let ctx_compute_trait_impl_name (ctx : extraction_ctx) (trait_decl : trait_decl)
   let name =
     let params = trait_impl.llbc_generics in
     let args = trait_impl.llbc_impl_trait.decl_generics in
-    let name = ctx_prepare_name trait_impl.span ctx trait_decl.llbc_name in
+    let name =
+      ctx_prepare_name trait_impl.item_meta.span ctx trait_decl.llbc_name
+    in
+    (* let name = rename_llbc_name trait_impl.item_meta name in  *)
```
sonmarcho, line 1552: Commented code.
You are also ignoring the `rename` attribute here.

compiler/ExtractBase.ml
```diff
@@ -1970,20 +1984,20 @@ let ctx_add_generic_params (span : Meta.span) (current_def_name : Types.name)
 let ctx_add_decreases_proof (def : fun_decl) (ctx : extraction_ctx) :
     extraction_ctx =
   let name =
-    ctx_compute_decreases_proof_name def.span ctx def.def_id def.llbc_name
-      def.num_loops def.loop_id
+    ctx_compute_decreases_proof_name def.item_meta.span ctx def.def_id
```
sonmarcho, line 1987: I think you should also use the `rename` attribute in `ctx_compute_decreases_proof_name`

compiler/ExtractBase.ml
```diff
@@ -1970,20 +1984,20 @@ let ctx_add_generic_params (span : Meta.span) (current_def_name : Types.name)
...
+  ctx_add def.item_meta.span
     (DecreasesProofId (FRegular def.def_id, def.loop_id))
     name ctx
 
 let ctx_add_termination_measure (def : fun_decl) (ctx : extraction_ctx) :
     extraction_ctx =
   let name =
-    ctx_compute_termination_measure_name def.span ctx def.def_id def.llbc_name
-      def.num_loops def.loop_id
+    ctx_compute_termination_measure_name def.item_meta.span ctx def.def_id
```
sonmarcho, line 1997: Same remark as for `ctx_compute_decreases_proof_name`

sonmarcho (changes requested):
There are minor mistakes.
Also, currently missing:
- rename variant names
- some rigorous tests

For the test: F* will need tests with the `-decreases-clauses` option (in addition to the other tests).

compiler/ExtractBase.ml
sonmarcho: Why have you marked this conversation as revolved? The issue is still there...

compiler/ExtractBase.ml
```diff
@@ -2053,4 +2055,4 @@ let ctx_add_fun_decl (def : fun_decl) (ctx : extraction_ctx) : extraction_ctx =
 let ctx_compute_type_decl_name (ctx : extraction_ctx) (def : type_decl) : string
     =
   let llbc_name = rename_llbc_name def.item_meta def.llbc_name in
```
sonmarcho, line 2057: I think this should be moved to `ctx_compute_type_decl_name_aux`, otherwise you will forget to update the LLBC name in other places

compiler/ExtractBase.ml
```diff
@@ -1417,9 +1417,9 @@ let ctx_compute_type_decl_name (span : Meta.span) (ctx : extraction_ctx)
     access then causes trouble because not all provers accept syntax like
     [x.3] where [x] is a tuple.
  *)
-let ctx_compute_field_name (def : type_decl) (ctx : extraction_ctx)
-    (def_name : llbc_name) (field_id : FieldId.id) (field_name : string option)
-    : string =
+let ctx_compute_field_name (def : type_decl) (item_meta : Meta.item_meta)
```
sonmarcho, line 1420:
```suggestion
let ctx_compute_field_name (def : type_decl) (field_meta : Meta.item_meta)
```

compiler/ExtractTypes.ml
```diff
@@ -822,9 +822,13 @@ let extract_type_decl_register_names (ctx : extraction_ctx) (def : type_decl) :
             | None ->
                 VariantId.mapi
                   (fun variant_id (variant : variant) ->
-                    let name =
+                    let variant_name = 
+                      (match variant.item_meta.rename with
+                        | Some(name) -> name
```
sonmarcho: The formatting is weird (`Some(name)`). Do you reformat the code before pushing?
@Nadrieril the CI should check that the code is well-formatted correct? Because I don't see the nix CI failing for this one.

compiler/ExtractTypes.ml
```diff
@@ -822,9 +822,13 @@ let extract_type_decl_register_names (ctx : extraction_ctx) (def : type_decl) :
             | None ->
                 VariantId.mapi
                   (fun variant_id (variant : variant) ->
-                    let name =
+                    let variant_name = 
+                      (match variant.item_meta.rename with
```
sonmarcho: You can use `Option.value`

compiler/ExtractBase.ml
sonmarcho: You are still ignoring the `rename` attribute here. Please be careful when marking conversations as resolved.

sonmarcho (changes requested):

tests/coq/arrays/Arrays.v
```diff
@@ -376,7 +376,7 @@ Definition non_copyable_array : result unit :=
 .
 
 (** [arrays::sum]: loop 0:
-    Source: 'tests/src/arrays.rs', lines 245:0-253:1 *)
+    Source: 'tests/src/arrays.rs', lines 245:0-245:28 *)
```
Nadrieril, line 379: The span changed. It used to point to [the whole function](https://github.com/AeneasVerif/aeneas/blob/4d3778bea3112168645efc03308056ec341abb5f/tests/src/arrays.rs#L245-L253) and now only points to the function signature. Is that intended and/or what we want?

tests/lean/RenameAttribute.lean
```diff
@@ -0,0 +1,84 @@
...
+
+/- Trait implementation: [rename_attribute::{(rename_attribute::BoolTrait for bool)}]
+   Source: 'tests/src/rename_attribute.rs', lines 20:0-20:23 -/
+def BoolImplBool : BoolTest Bool := {
+  getTest := BoolTraitBool.get_bool
+}
+
+/- [rename_attribute::BoolTrait::ret_true]:
+   Source: 'tests/src/rename_attribute.rs', lines 14:4-14:30 -/
+def BoolTrait.retTest
```
Nadrieril, line 26: Shouldn't this be called `BoolTest.retTest`?

tests/src/rename_attribute.rs
```diff
@@ -0,0 +1,74 @@
...
+struct Foo {
+    #[charon::rename("FieldTest")]
+    field1: u32,
+}
+
+#[charon::rename("Const_Test")]
+const C: u32 = 100 + 10 + 1;
+
+#[aeneas::rename("Type-Aeneas36")]
+type Test2 = u32;
```
Nadrieril, line 52: Aeneas ignores type aliases so this doesn't test anything

tests/coq/arrays/Arrays.v
sonmarcho: It should indeed only point to the loop body (so now is better)

tests/lean/Arrays.lean
```diff
@@ -349,7 +349,7 @@ def non_copyable_array : Result Unit :=
   take_array_t (Array.make AB 2#usize [ AB.A, AB.B ])
 
 /- [arrays::sum]: loop 0:
-   Source: 'tests/src/arrays.rs', lines 247:4-253:1 -/
+   Source: 'tests/src/arrays.rs', lines 245:0-245:28 -/
```
Nadrieril, line 352: That's surprising: looks like you're using the span of the function signature now instead of the span of the body?

tests/coq/arrays/Arrays.v
sonmarcho: That was actually an issue - I solved it

PR #240, opened 2024-06-12
backends/lean: introduce `HasIntPred` automation

RaitoBezarius:
`HasIntPred` enable generation of facts based on specific terms in the context rather than their types, e.g. if the "length of a list" occurs in the context, generate the fact 0 ≤ length of that list, which can be further used for `scalar_tac` automation to discharge bounds goals.

The aim is to use it to simplify various height related computations, e.g. whenever "height of a (left ; right) tree" is encountered, generate "height left < height of a (left ; right) tree", etc.


RaitoBezarius:
cc @sonmarcho

PR #244, opened 2024-06-14
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/254


sonmarcho:
We should really find a way of not displaying the commit hashes when printing spans: this creates too much noise.

Nadrieril:
Indeed. Filed as https://github.com/AeneasVerif/charon/issues/256.

PR #258, opened 2024-06-24
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/266

### DIFF
--- a/compiler/Pure.ml
+++ b/compiler/Pure.ml
@@ -355,14 +355,14 @@ and trait_instance_id =
 type field = {
   field_name : string option;
   field_ty : ty;
-  item_meta : item_meta;
+  attr_info : attr_info;
 }
 [@@deriving show]
 
 type variant = {
   variant_name : string;
   fields : field list;
-  item_meta : item_meta;
+  attr_info : attr_info;
 }
 [@@deriving show]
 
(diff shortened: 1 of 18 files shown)


compiler/Pure.ml
```diff
@@ -397,15 +397,14 @@ type predicates = { trait_type_constraints : trait_type_constraint list }
 
 type type_decl = {
   def_id : TypeDeclId.id;
-  is_local : bool;
   llbc_name : llbc_name;
       (** The original name coming from the LLBC declaration *)
```
sonmarcho, line 401: This is not necessary anymore, because the name is now stored in the `item_meta` field (for which we should add a comment). Same for the other declarations.

Nadrieril: That's not true because of renames. Or do you want renames to update `item_meta`?

sonmarcho: The LLBC name is kept inside the `item_meta`, together with the `rename` attribute. When we generate the name for the extracted definitions we combine the LLBC name (today, from this field, but could be from the `item_meta`) with the content of the rename attribute, if there is.

Nadrieril: Oh, I thought that was done earlier in the pipeline. Makes sense to remove the field then

Nadrieril: Done here: https://github.com/AeneasVerif/aeneas/pull/272

PR #259, opened 2024-06-24
Use the self-hosted runner for the lean CI checks

Nadrieril:
The runner is faster and we can reuse the nix cache.


Nadrieril:
(It is indeed 3x as fast)

sonmarcho:
That is really good, thanks!

PR #264, opened 2024-06-25
Tweak formatting options

Nadrieril:


sonmarcho (approved):
Looks good to me thanks! Let's use those formatting options for now, and maybe tweak them further more later.

PR #277, opened 2024-07-01
feat: progress? to say which theorem has been used

RaitoBezarius:
Like `simp?` and similar, `progress?` will provide what theorem was used to make the goal progress.

TODO: 

- [x] grab original stx elements
- [x] show `progress ...` instead of `...` during the print.
- [x] make code action work on this fragment


backends/lean/Base/Progress/Progress.lean
```diff
@@ -370,14 +373,22 @@ def evalProgress (args : TSyntax `Progress.progressArgs) : TacticM Unit := do
       Arith.scalarTac false
     else
       throwError "Not a linear arithmetic goal"
-  progressAsmsOrLookupTheorem keep withArg ids splitPost (
+  let usedTheorem ← progressAsmsOrLookupTheorem keep withArg ids splitPost (
     withMainContext do
     trace[Progress] "trying to solve assumption: {← getMainGoal}"
     firstTac [assumptionTac, scalarTac])
-  trace[Diverge] "Progress done"
```
RaitoBezarius, line 377: @sonmarcho here, I think it was meant to be Progress, I renamed it even though it has nothing to do with this PR, is this fine or should I split the change?

sonmarcho: Yes, that's fine, thanks! :)

backends/lean/Base/Progress/Progress.lean
```diff
@@ -370,14 +376,26 @@ def evalProgress (args : TSyntax `Progress.progressArgs) : TacticM Unit := do
...
+  trace[Progress] "Progress done"
+  return {
+    usedTheorem := usedTheorem
```
sonmarcho, line 385:
```suggestion
    usedTheorem
```

sonmarcho: This is correct Lean code right?

sonmarcho (approved):
This looks good! I have a minor suggestion, but this put aside you're good to merge :)

sonmarcho:
However it seems that it reordered some definitions, leading to some diff in the files generated by Aeneas: this is slightly suspicious.

backends/lean/Base/Progress/Progress.lean
RaitoBezarius: Done, working!

RaitoBezarius:
@sonmarcho feel free to merge :)

PR #278, opened 2024-07-03
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/282


sonmarcho (commented):
I'm not happy with the fact that some definitions moved: it shouldn't happen.

Nadrieril:
Could it be that the order of traversal in llbc is different from ullbc?

PR #281, opened 2024-07-03
feat: progress with a term

RaitoBezarius:
`progress with X` where X can now be a full term, e.g. `X := (@SomeSpec.XYZ T U)`.

TODO:

- [ ] diagnose what happens with `as` and some failures
- [ ] add unit tests


RaitoBezarius:
@sonmarcho I think this is almost ready, there might be still some vague TODOs, but it feels feature-complete, I will take it for a walk and try to come up with proper interesting unit tests.

The case:
```
progress with (_: T)
```

fails because Lean complains about inability to synthesize the placeholder, I wonder if I should try to fix it in the code by modifying the calls to synthesization and allow failure and create for each failure a new metavar and leave it as a new goal to be assigned by the user (which I feel like is the solution to fix that usecase).

sonmarcho:
> @sonmarcho I think this is almost ready, there might be still some vague TODOs, but it feels feature-complete, I will take it for a walk and try to come up with proper interesting unit tests.
> 
> The case:
> 
> ```
> progress with (_: T)
> ```
> 
> fails because Lean complains about inability to synthesize the placeholder, I wonder if I should try to fix it in the code by modifying the calls to synthesization and allow failure and create for each failure a new metavar and leave it as a new goal to be assigned by the user (which I feel like is the solution to fix that usecase).

We might do something like this in the future. I don't see the point of doing this for regular function calls, because then you should always call `progress` with a precise lemma in mind (or an instantiated lemma, hence the need for terms instead of ids). However, we might want to give the possibility of writing a specification when considering `if then else` expressions like below, where we would open two goals: one to prove that the expression in the let-binding has the proper spec, and another for after we evaluated the `if then else`.
```lean
let x <- if b then 0 else 1 -- We can't use a lemma here
let y <- f x
```

backends/lean/Base/Progress/Progress.lean
```diff
@@ -219,7 +209,7 @@ def getFirstArg (args : Array Expr) : Option Expr := do
    Return true if it succeeded. -/
 def tryLookupApply (keep : Option Name) (ids : Array (Option Name)) (splitPost : Bool)
```
sonmarcho, line 210: I think we should rename this function to `tryApply`, because it doesn't perform a lookup anymore.
Also, it might make sense to move the match over the option (and the comments printed in the trace are not valid anymore I believe).

backends/lean/Base/Progress/Progress.lean
```diff
@@ -343,17 +335,27 @@ def evalProgress (args : TSyntax `Progress.progressArgs) : TacticM Stats := do
...
-        -- Not a local declaration: should be a theorem
-        trace[Progress] "With arg: theorem"
-        addCompletionInfo <| CompletionInfo.id id id.getId (danglingDot := false) {} none
-        let some (.const name _) ← Term.resolveId? id | throwError m!"Could not find theorem: {id}"
-        pure (some (.Theorem name))
+
+      -- Either, it's a local declaration or a theorem, either it's a term
+      -- for which we need to find a declaration
+
+      if id.isIdent then
```
sonmarcho, line 342: I don't think we really need this case disjunction. Shouldn't it be performed by Lean's elaboration (i.e., isn't the `else` branch enough)?

backends/lean/Base/Progress/Progress.lean
```diff
@@ -392,9 +394,12 @@ elab tk:"progress?" args:progressArgs : tactic => do
   let stats ← evalProgress args
   let mut stxArgs := args.raw
   if stxArgs[1].isNone then
-    let withArg := mkNullNode #[mkAtom "with", mkIdent stats.usedTheorem]
+    let withArg := mkNullNode #[mkAtom "with", stats.usedTheorem]
     stxArgs := stxArgs.setArg 1 withArg
   let tac := mkNode `Progress.progress #[mkAtom "progress", stxArgs]
+  -- if stats.usedTheorem.isInaccesisble then
+  --  logInfo "{...} is inaccessible, if you want to actually use the suggestion, please use `rename_i ...`"
```
sonmarcho, line 401: Please remove all commented code (note however that printing such a comment would be useful, if it is possible).

backends/lean/Base/Progress/Progress.lean
RaitoBezarius: Hm, if I do that, something like:
```
set_option trace.Progress true in
  example {x: U32}
    (f : Result U32 → Result Unit)
    (h : ∀ (x: U32), ∀ (y: U32), f (x + y) = .ok ()):
      f (.ok x) = ok () := by
      have : ∀ x, f x = .ok () := by sorry
      progress with this
```

won't work because `this` elaborated := `sorryAx ?m.something true` and the metavariable will create an obscure failure of progress here.

RaitoBezarius: (tbh, the current code does not work neither, it fails by not finding `this` in the context)

sonmarcho: Ok, this is good to know (and worth commenting in the code ;) )

backends/lean/Base/Progress/Progress.lean
```diff
@@ -268,17 +258,17 @@ def progressAsmsOrLookupTheorem (keep : Option Name) (withTh : Option TheoremOrL
   match withTh with
   | some th => do
     match ← progressWith fExpr th keep ids splitPost asmTac with
-    | .Ok => return th
+    | .Ok => return (← exprToSyntax th)
```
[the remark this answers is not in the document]
sonmarcho: Does it generate spurious syntax?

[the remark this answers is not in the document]
RaitoBezarius: Yeah, just dump a metavariable I think

backends/lean/Base/Progress/Progress.lean
RaitoBezarius: Wouldn't this be a blocker though for the PR? I'm trying to make this work but I'm just diving into how the whole thing is supposed to work

sonmarcho: What I meant is that the fact that we need the case disjunction is good to know and should be documented. But of course, we need to make the elaboration work, generally speaking.

backends/lean/Base/Progress/Progress.lean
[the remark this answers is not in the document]
sonmarcho: I'm not sure I understand in which situations it dumps a metavariable

PR #285, opened 2024-07-08
doc: initialize a user-facing mdbook

RaitoBezarius:
@sonmarcho @R1kM Here's my proposal regarding documenting `progress` among other things that I learned during the FV of AVL trees.

A mdBook-style user facing documentation with a part on "Proving Rust code with Lean" which dumps all that I know about using Aeneas & Lean in practice on Rust code.

Question is: do I do everything in this PR or shall we just get the CI right and improve on it incrementally as we commit to an outline, etc. ?


sonmarcho:
Thanks for starting this, it was much needed!
Following our conversation, we should first get the CI right and do things incrementally. In particular, once we get the outline and the CI working, anybody can start making changes.

RaitoBezarius:
Moved to an Alectryon setup.

sonmarcho:
@Nadrieril I think it's important that you keep an eye on this PR as well.

RaitoBezarius:
@Nadrieril ping on this :)

.github/workflows/build-userdocs.yml
```diff
@@ -0,0 +1,14 @@
...
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Test Aeneas user docs
+    runs-on: [self-hosted, linux, nix]
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
Nadrieril, line 11: Why this? By default the clone is shallow, this makes it deep. I don't think we need that

Nadrieril (approved):

.github/workflows/build-userdocs.yml
RaitoBezarius: for building, yes, i'm dumb

RaitoBezarius: Fixed.

sonmarcho:
The PR https://github.com/AeneasVerif/aeneas/pull/282 is blocked because the doc building fails.

RaitoBezarius:
Urgh.

RaitoBezarius:
Sent a hotfix in another PR.

PR #294, opened 2024-08-10
docs: init first sketch

RaitoBezarius:
This is a first sketch of user documentation for Aeneas to prove things with Lean.

Maybe "height" is too much.

Here's some details about draft chapters:

- Extracting to Lean: review the Lean extraction of Aeneas and relate Rust code to Lean code.
- Panic-freedomness: talk about how Aeneas will extract a bunch of things with `Result X`, explain why that happens and why the first objective is usually to lift that `Result X` into a proper `X` and determine the preconditions that will make this happen all the time.
- Verifying `Find` operation: focus on the `find` extraction in Lean, explain the structure of the verification (a `find_loop_spec` and a `find_spec`) and the required theory on the side for it.
- What does it mean to be in a tree?: modelling the set of values of a functional tree as a set-theoretic object, i.e. a `T → bool` recursive predicate that says `x = a \/ x \in left.set \/ x \in right.set`, compare with the `ForallNode` definition, suggest the existence of an alternative inductive definition of `MemTree`, explains the tradeoffs with respect to reuse of the mathlib library.
- Applying Known Specifications?  Introduce to `progress` and show how to use it in the `find` verification.
- _uncertain_ Dealing with Loops? Explain that a specification can call itself recursively and show examples where the recursive call can be non-trivial.
- Verifying Insert Operations? focus on `insert` extraction in Lean, explain the structure of the verification, required theory on the side of it, reusing parts of the `find` proof.
- _uncertain_ Dealing with Large Proofs? Showing how to diagnose long proofs, how to split them up, profiling, etc.
- _uncertain_ Dealing with Symmetry? Showing tricks to merge goals together after shuffling them, etc.


Makefile
```diff
@@ -112,6 +112,8 @@ format:
 	cd tests/test_runner && dune fmt || true
 	rustfmt $(RUSTFMT_FLAGS) $(INPUTS_DIR)/*.rs
 	cd $(INPUTS_DIR)/betree && cargo fmt $(RUSTFMT_FLAGS)
+	cd $(INPUTS_DIR)/naive-tree && cargo fmt $(RUSTFMT_FLAGS)
+	cd $(INPUTS_DIR)/avl && cargo fmt $(RUSTFMT_FLAGS)
```
sonmarcho, line 116: I think we should have one generic command to format all the subdirectories we find in `tests/src` (I can do that, I did it in other Makefiles)

Makefile
```diff
@@ -138,13 +140,19 @@ clean-generated-llbc:
...
 .PHONY: betree-tests
 betree-tests:
 	cd $(INPUTS_DIR)/betree && $(MAKE) test
 
+avl-tests:
+	cd $(INPUTS_DIR)/avl && $(MAKE) test
+
+naivetree-tests:
+	cd $(INPUTS_DIR)/naive-tree && $(MAKE) test
+
```
sonmarcho, line 155: Same remark as for formatting

scripts/update-charon-pin.sh
```diff
@@ -10,11 +10,11 @@ if [ -L charon ]; then
     nix flake lock --override-input charon "github:aeneasverif/charon/$COMMIT"
 else
     echo 'Pinning the latest commit from Charon `main`'
-    nix flake lock --update-input charon
+    nix flake update charon
 fi
 # In both cases we also update the rust overlay to make sure we can get the
 # latest rust nightlies.
-nix flake lock --update-input charon/rust-overlay
+nix flake update charon/rust-overlay
```
sonmarcho, line 17: I will leave the review of this to Guillaume :)

Makefile
RaitoBezarius: OK, I will leave those commits out then and leave it to you!

Makefile
RaitoBezarius: Will do the same as for formatting.

scripts/update-charon-pin.sh
RaitoBezarius: @Nadrieril I think it may not work on Nix 2.18, but it broke on more modern Nix, so not sure what to do. I guess command testing makes more sense here.

Nadrieril: I've used that before on 2.18, why may this not work?

RaitoBezarius: I'm pretty sure there's a Nix version that doesn't support `nix flake update XYZ`, I thought it was 2.18 but I guess I'm wrong.

Nadrieril: ah, it does not look like it does what I want indeed. `nix flake update rust-overlay` on charon fails with `cannot find flake 'flake:rust-overlay' in the flake registries` which is not what I want

RaitoBezarius: I can drop this part of the PR, tbh. This was more chores that I picked up as I went trying to update my repo.

RaitoBezarius: Dropped!

RaitoBezarius:
PTAL @sonmarcho I dropped the irrelevant parts.

RaitoBezarius:
@sonmarcho In principle, I updated to 4.11.0-rc2, let's see how it goes.

RaitoBezarius:
@pnmadelaine We will need a cache for the `userdocs` part because it will compile Mathlib4 in Nix and if we do it at each rerun, this will be very expensive.

sonmarcho:
This is great, thanks!! I will review and merge asap.

docs/user/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+The `result/` symlink is a HTML webroot, this can be deployed as a static website.
+
+## My build is failing due to a compilation error but it works outside of Nix
+
+This is highly likely due to a mismatch between the dependencies in the Nix setup and outside.
+
+The Nix setup reads the `lakefile-manifest.json`, but this lockfile is insufficient to lock purely the dependencies (NAR SHA256 serialization are missing).
+
+To overcome this, they are specified manually in `flake.nix`, but each time the Aeneas standard library updates their dependencies, the `flake.nix` needs to have all its hash rewritten.
```
sonmarcho, line 82: I don't fully understand what we are supposed to do when updating the dependencies, then

docs/user/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+### How to add a new Lean dependency to the project?
+
+Each `buildLeanPackage` can take a list of `deps`.
+
+Two cases:
+
+- either, this is a local dependency and this can be built like the `Base` (Aeneas standard library) one.
+- either, this is a remote dependency and this is often added to a Lakefile, this can be built like `Mathlib` and the others.
```
sonmarcho, line 64:
```suggestion
- or, this is a remote dependency and this is often added to a Lakefile, this can be built like `Mathlib` and the others.
```

docs/user/alectryon.js
```diff
@@ -0,0 +1,190 @@
+var Alectryon;
```
sonmarcho, line 1: This file also comes from Alectryon right? There are no copyrights (that's surprising)?

sonmarcho (commented):
This is great, thanks!
I put a few questions, and am also wondering: you mention that the files to process should be put in `lean/basics/`. But what about the other folders? For instance `lean/bst`?

docs/user/alectryon.js
RaitoBezarius: https://github.com/cpitclaudel/alectryon/blob/2a544230290be6dd84f0ce736e86f8a071d70739/alectryon/assets/alectryon.js#L4

Yep, but this is how it is upstream, I am fine with adding the compliance copyright, what do you prefer?

RaitoBezarius: (may be worth investigating https://reuse.software/ if we want to maximize the licensing compliance situation)

RaitoBezarius:
> This is great, thanks! I put a few questions, and am also wondering: you mention that the files to process should be put in `lean/basics/`. But what about the other folders? For instance `lean/bst`?

I updated the docs, but there's a section on this.
Other directories that are not listed as roots of the `literate` Lean meta package does not get their Lean files processed.

docs/user/README.md
RaitoBezarius: I expanded much more the documentation, let me know how you feel about it now

PR #297, opened 2024-08-14
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/308

### DIFF
--- a/compiler/SymbolicToPure.ml
+++ b/compiler/SymbolicToPure.ml
@@ -578,7 +578,7 @@ let translate_type_decl_kind (span : Meta.span) (kind : T.type_decl_kind) :
   | Alias _ ->
       craise __FILE__ __LINE__ span
         "type aliases should have been removed earlier"
-  | T.Opaque -> Opaque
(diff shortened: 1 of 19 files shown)


compiler/SymbolicToPure.ml
```diff
@@ -578,7 +578,7 @@ let translate_type_decl_kind (span : Meta.span) (kind : T.type_decl_kind) :
   | Alias _ ->
       craise __FILE__ __LINE__ span
         "type aliases should have been removed earlier"
-  | T.Opaque -> Opaque
+  | T.Opaque | T.Error _ -> Opaque
```
sonmarcho, line 581: We should raise an exception in the Error case

Nadrieril: The exception has already been raised by charon. If we're still here we may as well process as much as possible.

Nadrieril: Btw this can only happen with `charon --errors-as-warnings`

sonmarcho: Ok, makes sense

PR #320, opened 2024-08-27
test(bst): add BST as an example

RaitoBezarius:
This adds BST as an example from my internship's staging repository.

It adds a bunch of unit tests for them.


sonmarcho:
They're simply binary trees, right? Maybe "binary tree" is a better name than "naive tree"?...

RaitoBezarius:
Right, I don't have a strong opinion because AVL trees are binary trees as well.

RaitoBezarius:
PTAL @R1kM @sonmarcho, I renamed it.

PR #328, opened 2024-08-30
chore(lean): minimize imports in `Base/Arith`

RaitoBezarius:
This is a methodology employed by the Mathlib community to improve the build performance of Mathlib.

The idea is to exploit the ImportGraph library which has a command `#min_imports` and `#redundant_imports` to figure out what imports to trim down.

See: https://reservoir.lean-lang.org/@leanprover-community/importGraph for the docs.

This is the result of this process for `Base/Arith`.

@sonmarcho If you are happy with the idea, I can minimize the rest of the library. This will be helpful with the documentation PR to minimize the build times as well (as we won't build unnecessary Mathlib modules in Nix).

### DIFF
--- a/backends/lean/Base/Arith/Base.lean
+++ b/backends/lean/Base/Arith/Base.lean
@@ -1,5 +1,5 @@
-import Lean
-import Mathlib.Tactic.Linarith -- Introduces a lot of useful lemmas
+import Mathlib.Algebra.Order.Ring.Int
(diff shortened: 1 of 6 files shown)


sonmarcho:
I think this is a great idea! You don't have to do this for the whole library, but it would be good to document this somewhere, perhaps in the README or in a CONTRIBUTING.md file ?

PR #332, opened 2024-09-06
Propagate Float literal changes from Charon

R1kM:
Companion PR to https://github.com/AeneasVerif/charon/pull/348


Nadrieril:
You didn't update the charon pin and CI was red, how come this was merged?

Issue #346, opened 2024-11-06
Add saturating add/sub to the standard library


sonmarcho:
See for instance: https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub

sonmarcho:
Solved by https://github.com/AeneasVerif/aeneas/pull/348

PR #348, opened 2024-11-06
Mark saturating_add/sub as a builtin, and add Lean models to Primitives

R1kM:
Extend the Rust translation with support for the `saturating_add` and `saturating_sub` builtins.
This also includes a few lemmas annotated with `pspec` for use with progress for unsigned saturating additions.

There are a few surprising parts in the Lean proofs, that might be solved by using omega instead of linarith in scalar_tac

### DIFF
--- a/backends/lean/Base/Arith/Int.lean
+++ b/backends/lean/Base/Arith/Int.lean
@@ -121,9 +121,9 @@ def intTacPreprocess (extraPrePreprocess extraPreprocess :  Tactic.TacticM Unit)
   -- Pre-preprocessing
   extraPrePreprocess
   -- Apply the forward rules
-  intTacSaturateForward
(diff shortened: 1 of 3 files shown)


R1kM:
After some debugging with Son, it turns out that omega is already enabled, but that the unexpected behavior of scalar_tac was due to some goals being solved during preprocessing, and the tactic failing in later stages due to the absence of goals.
[962c695](https://github.com/AeneasVerif/aeneas/pull/348/commits/962c695ad2fdad2e08daa55391d277c8abf3fb05) attempts to solve this.

sonmarcho:
This solves https://github.com/AeneasVerif/aeneas/issues/346

sonmarcho (approved):

PR #353, opened 2024-11-10
Implement a micro-pass to simplify the use of the array/slice index functions and the aggregated ADTs

sonmarcho:
# Arrays
Today, whenever we update an element in an array or a slice (e.g., `x[0] = 1`) we first create a mutable borrow to the element of the array, then use this borrow to update the element (e.g., `let tmp = &mut x[0]; *tmp = 1`). The output code looks like this:
```ocaml
let (_, back) = Array.index_mut a i in
let a' = back v in
...
```

I implemented a micro-pass so that the output code is actually:
```ocaml
let a' = Array.update a i v in
...
```

# Aggregated ADTs
Because of the way the symbolic execution works, we also often see patterns of the following shape in the generated code:
```lean
if x.field then { x with field = true }
else { x with field = false }
```
This PR implements a micro-pass to simplify this to:
```lean
if x.field then x
else x
```

This is very cosmetic but makes the generated code a lot cleaner and understandable in some cases.

### DIFF
--- a/compiler/Extract.ml
+++ b/compiler/Extract.ml
@@ -589,6 +589,14 @@ and extract_function_call (span : Meta.span) (ctx : extraction_ctx)
         | FromLlbc (FunId (FAssumed aid), _) ->
             Some
               (Assumed.AssumedFunIdMap.find aid ctx.builtin_sigs).explicit_info
+        | Pure (UpdateAtIndex Array) ->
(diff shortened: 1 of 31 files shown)


sonmarcho:
This PR relies on https://github.com/AeneasVerif/charon/pull/456

PR #357, opened 2024-11-13
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/457.

There's a tiny bit of trickiness around how `FailBottom` remembers how deep in the place we are so that later we can recover the subplace. It might make sense to store the whole subplace instead of the depth.


compiler/SynthesizeSymbolic.ml
```diff
@@ -8,8 +8,17 @@ open Errors
...
     =
-  let bv = Contexts.ctx_lookup_var_binder span ctx p.var_id in
-  { bv; projection = p.projection }
+  let rec place_to_projection_and_var (place : place) :
+      var_id * projection_elem list =
+    match place.kind with
+    | PlaceBase var_id -> (var_id, [])
+    | PlaceProjection (subplace, pe) ->
+        let var_id, proj = place_to_projection_and_var subplace in
+        (var_id, pe :: proj)
```
sonmarcho, line 17: We should compute the list of projection elements in the reverse order.
Also, I'm going to check if we really need to store this information.

sonmarcho (commented):
This looks good to me. I'm going to make a few cleaning changes and approve the PR once they are done.

compiler/SynthesizeSymbolic.ml
Nadrieril: oh oops, you're right

sonmarcho (approved):

PR #361, opened 2024-11-18
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/461


src/Translate.ml
```diff
@@ -631,10 +631,13 @@ let export_functions_group_scc (fmt : Format.formatter) (config : gen_config)
...
-        (* Check if the definition needs to be filtered or not *)
+        (* Check if the definition needs to be filtered or not. Functions that
+           are the initializers of globals are processed when we process the
+           corresponding global. *)
         if
-          ((not is_opaque) && config.extract_transparent)
-          || (is_opaque && config.extract_opaque)
+          (((not is_opaque) && config.extract_transparent)
+          || (is_opaque && config.extract_opaque))
+          && not def.is_global_decl_body
```
Nadrieril, line 640: @sonmarcho this is the one place that the update wasn't obvious

sonmarcho: Ok I understand what you're doing.
I suggest the following. What I wrote above still stands: the role of this check is to filter the opaque/transparent functions. The fact that we also ignore the global bodies *here* is confusing. I suggest filtering *groups* (i.e., before calling `export_functions_group`) if the group contains a global initializer then we ignore the whole group, and check that this group doesn't contain functions which are not global initializers (and that it only contains a single non recursive definition I guess).

Nadrieril: I did that in the last commit, can you have a look?

Issue #75, opened 2024-02-22
Add support for ADTs with lifetimes


zhassan-aws:
Code:
```rust
struct Foo<'a> {
    x: &'a i32,
}

fn main() { }
```
Output:
```
$ charon
   Compiling ref_in_struct v0.1.0 (/home/ubuntu/examples/aeneas/ref_in_struct)
[ INFO charon_driver::export:106] [gexport]: Generated the file: /home/ubuntu/examples/aeneas/ref_in_struct/ref_in_struct.llbc
    Finished release [optimized] target(s) in 0.15s
$ aeneas -backend lean ref_in_struct.llbc 
[Info ] Imported: ref_in_struct.llbc
Uncaught exception:
  
  "Assert_failure SymbolicToPure.ml:554:2"

Raised at Aeneas__SymbolicToPure.translate_type_decl in file "SymbolicToPure.ml", line 554, characters 2-23
Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
Called from Aeneas__Translate.translate_crate_to_pure in file "Translate.ml", line 277, characters 19-64
Called from Aeneas__Translate.translate_crate in file "Translate.ml", line 992, characters 4-33
Called from Dune__exe__Main in file "Main.ml", line 279, characters 6-58
```

sonmarcho:
Yes, we don't support ADTs with lifetimes for now (I intend to lift this soon).
Also, on the shorter term: we're about to start an overhaul of Aeneas to have informative error messages (I've been wanting this for a while and I think it's time).

zhassan-aws:
This also impacts `std::str::chars`, e.g:
```rust
fn foo(s: &str) {
    let _c = s.chars();
}
```
```
$ aeneas -backend lean ref_in_struct.llbc 
[Info ] Imported: ref_in_struct.llbc
Uncaught exception:
  
  Not_found

Raised at Stdlib__Map.Make.find in file "map.ml", line 137, characters 10-25
Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
Called from Aeneas__Translate.export_types_group in file "Translate.ml", line 415, characters 4-72
Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15
Called from Aeneas__Translate.extract_definitions in file "Translate.ml", line 822, characters 2-52
Called from Aeneas__Translate.extract_file in file "Translate.ml", line 940, characters 2-36
[...]

PR #367, opened 2024-11-26
Add support for structures containing borrows

sonmarcho:
This PR adds supports for structures containing borrows. Enumerations with borrows will come later: I need to update the join operation to make this work properly. This PR needs https://github.com/AeneasVerif/charon/pull/475 on the Charon side.


sonmarcho:
An important design choice : if a function consumes a structure containing mutable borrows for, say, some lifetime `'a`, then the backward function returns the whole structure, where only the fields for the mutable borrows have been updated.
For instance, I put this in the tests:
```rust
struct MutWrapper2<'a, 'b, T> {
    x: &'a mut T,
    y: &'b mut T,
}

impl<'a, 'b, T> MutWrapper2<'a, 'b, T> {
    fn unwrap(self: Self) -> (&'a mut T, &'b mut T) {
        (self.x, self.y)
    }
}
```

The generated model is as follows:
```lean
structure MutWrapper2 (T : Type) where
  x : T
  y : T

def MutWrapper2.unwrap
  {T : Type} (self : MutWrapper2 T) :
  Result ((T × T) × (T → MutWrapper2 T) × (T → MutWrapper2 T))
  :=
  let back'a := fun ret => { self with x := ret }
  let back'b := fun ret => { self with y := ret }
  Result.ok ((self.x, self.y), back'a, back'b)
```

This might seem like a strange choice, as we could also generate the following model:
```lean
def MutWrapper2.unwrap
  {T : Type} (self : MutWrapper2 T) :
  Result ((T × T) × (T → T) × (T → T))
  :=
  let back'a := fun ret => ret
  let back'b := fun ret => ret
  Result.ok ((self.x, self.y), back'a, back'b)
```
(i.e., the backward functions give back exactly the tuple of updated fields)

I made this choice because it is not possible to do so for (recursive) enumerations, unless we introduce auxiliary definitions (one auxiliary definition for each lifetime which is used by mutable borrows in the enumeration, to "collect" exactly the values that need to be given back), and doing this would be extremely cumbersome. In prevision of the future support for enumerations with borrows, I decided to make the treatment of ADTs consistent (I do not want to differentiate structures from enumerations at this point).

sonmarcho:
Also, an important note: the type analysis I do to compute what backward functions consume/produce is unsound if indirections introduced by trait impls are not removed (this [PR](https://github.com/AeneasVerif/charon/pull/210) - this is not a fundamental problem, I just need to put a bit more engineering time into properly handling this), and if there are constraints over associated types (which would be removed by [this](https://github.com/AeneasVerif/charon/issues/127)). For now I generate errors if I detect those, and I added an option to downgrade those errors to warnings, so that the `trait.rs` test goes through.

zhassan-aws:
Awesome! Does this resolve #75?

sonmarcho:
Yes it does! I'm marking it as "resolved"

PR #387, opened 2024-12-05
Clean up region variables

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/484. This propagates the changes from that PR, with the caveat that we rename `FreeRegionId` to `RegionId` because it's used everywhere.


src/interp/InterpreterUtils.ml
```diff
@@ -388,7 +388,7 @@ let compute_ids () =
         loan_ids := BorrowId.Set.add id !loan_ids
 
       method! visit_abstraction_id _ id = aids := AbstractionId.Set.add id !aids
-      method! visit_region_id _ id = rids := RegionId.Set.add id !rids
+      method! visit_free_region _ id = rids := RegionId.Set.add id !rids
```
sonmarcho, line 391: We should keep the `visit_region_id` (especially as some ids may not appear inside the `Free` variant but, for instance, inside a set in a region abstraction - that's actually a good reason for having a clear separation between free variables and bound variables).

src/llbc/Values.ml
```diff
@@ -221,7 +219,7 @@ class ['self] iter_typed_avalue_base =
       fun _ _ -> ()
 
     method visit_region_id_set : 'env -> region_id_set -> unit =
-      fun env ids -> RegionId.Set.iter (self#visit_region_id env) ids
+      fun env ids -> RegionId.Set.iter (self#visit_free_region env) ids
```
sonmarcho, line 222: We should keep `visit_region_id`

src/llbc/Values.ml
```diff
@@ -241,7 +239,7 @@ class ['self] map_typed_avalue_base =
       fun _ m -> m
 
     method visit_region_id_set : 'env -> region_id_set -> region_id_set =
-      fun env ids -> RegionId.Set.map (self#visit_region_id env) ids
+      fun env ids -> RegionId.Set.map (self#visit_free_region env) ids
```
sonmarcho, line 242: We should keep `visit_region_id`

src/interp/InterpreterUtils.ml
Nadrieril: oh yeah, I wrote this expecting we'd merge the ids but that's no longer needed

Nadrieril: Fixed, the drawback is that it's called `visit_free_region_id` because that's what charon calls it, whereas aeneas renames it to `region_id`. I could rename to `free_region_id` in aeneas too.

sonmarcho: Thanks! Is there a way of renaming it to `visit_region_id`?
The reason I would like to keep it that way is that in Aeneas regions are mostly used during the symbolic execution (whenever we call a function with lifetimes, we introduce fresh regions). In that context it is a bit weird to call them "free regions"...

Nadrieril: I "renamed" it in the last commit.

PR #407, opened 2025-01-02
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/512.

### DIFF
--- a/src/extract/Extract.ml
+++ b/src/extract/Extract.ml
@@ -2361,7 +2361,11 @@ let extract_trait_decl_method_names (ctx : extraction_ctx)
           in
           (item_name, name)
         in
-        List.map (fun (name, id) -> compute_item_name name id) required_methods
+        List.map
+          (fun (name, bound_fn) ->
+            let open Charon.Types in
+            compute_item_name name bound_fn.binder_value.fun_id)
+          required_methods
     | Some info ->
         (* This is a builtin *)
         let funs_map = StringMap.of_list info.methods in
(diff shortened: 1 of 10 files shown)


src/extract/Extract.ml
```diff
@@ -2785,13 +2791,16 @@ let extract_trait_decl_coq_arguments (ctx : extraction_ctx) (fmt : F.formatter)
...
+      let open Charon.Types in
+      let fun_id = bound_fn.binder_value.fun_id in
       (* Lookup the definition to retrieve the information about the explicit/implicit parameters *)
-      let trans = A.FunDeclId.Map.find id ctx.trans_funs in
+      let trans = A.FunDeclId.Map.find fun_id ctx.trans_funs in
       let f = trans.f in
       let explicit_info =
         explicit_info_drop_prefix decl.generics f.signature.explicit_info
       in
+      (* TODO: this is likely incorrect, we should instantiate the binders properly *)
```
Nadrieril, line 2803: @sonmarcho This will likely cause issues after https://github.com/AeneasVerif/charon/issues/127: the full method generics will no longer be a simple concatenation of the impl generics + method generics (because I'll be adding the extra associated type parameters at the end of the list of parameters instead of in the middle). We should find a way to instantiate the binders properly here.

src/interp/InterpreterStatements.ml
```diff
@@ -728,6 +724,7 @@ let eval_transparent_function_call_symbolic_inst (span : Meta.span)
                      we also need to update the generics.
                   *)
                   let func = FunId fid in
+                  (* TODO: the `trait_method_generics` look fishy *)
```
Nadrieril, line 727: @sonmarcho What are the `trait_method_generics` supposed to correspond to? From the name I'd expect these would be the generics _just for the method_, but here you pass generics valid for the function item (which include impl generics).

PR #411, opened 2025-01-07
Update the join for symbolic values with borrows

sonmarcho:
Following https://github.com/AeneasVerif/aeneas/pull/399, this PR updates the join operation and the computation of the fixed-points so that symbolic values with borrows are now supported with loops.


src/Main.ml
```diff
@@ -168,6 +173,12 @@ let () =
         Arg.Set print_unknown_externals,
         " Print all the external definitions which are not listed in the \
          builtin functions" );
+      ( "-log",
+        Arg.String add_activated_logger,
+        " Activate debugging log for a given logger designated by its name. \
+         The existing loggers are: {"
+        ^ String.concat ", " (Collections.StringMap.keys !loggers)
+        ^ "}" );
```
Nadrieril, line 181: Thanks for this, I'd wanted it for a while :pray:

sonmarcho: Yes, it had become really painful ^^

PR #412, opened 2025-01-07
Bind methods in pure

Nadrieril:
The last few Charon PRs made it so we can support cases where the generics for a method fn item aren't a simple concatenation of the trait_decl/trait_impl generics + the method generics. I hadn't yet propagated these changes into Pure; this PR does that. This happens to simplify the extraction of impl methods quite a bit because we now have the correct `explicit_info` available.

For simplicity I did not attempt to make the `de_bruijn_id`s correct in `Pure` since we don't need them to be. I commented the relevant places appropriately so we know what's happening if we end up needing more complex binding situations in the future.

### DIFF
--- a/src/extract/Extract.ml
+++ b/src/extract/Extract.ml
@@ -1364,7 +1364,7 @@ let extract_fun_parameters (space : bool ref) (ctx : extraction_ctx)
     This is used for opaque function declarations, in particular.
  *)
 let extract_fun_input_parameters_types (span : span) (ctx : extraction_ctx)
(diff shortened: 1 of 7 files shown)


sonmarcho (approved):
Looks good! Checking that I understand: the `generic_origin` is useful only when extracting trait impls, because there we *do* need two levels of bindings, one for the generics bound at the level of the trait impl, and another at the level of the item itself, like so:
```OCaml
struct FooImpl T { (* Item level *)
    f = fun U -> f0 T U (* Method level *)
}
```
Correct?

PR #413, opened 2025-01-08
Let charon manage `Self` clauses

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/514.

This removes the need for managing the special `Self` clause in methods ourselves. One consequence is that in https://github.com/AeneasVerif/charon/pull/514 I had to add a pass that removes unused `Self` clauses otherwise any trait impl that reuses an associated constant default would end up self-recursive. That's why you can see in the tests diff that some provided methods end up losing their `Self` clause: it's because of that same pass. This may be leaving the translation in a bit of an incorrect state while https://github.com/AeneasVerif/charon/issues/180 isn't finished.

### DIFF
--- a/src/pure/PureMicroPasses.ml
+++ b/src/pure/PureMicroPasses.ml
@@ -3622,7 +3622,7 @@ let add_type_annotations_to_fun_decl (trans_ctx : trans_ctx)
           (* Lookup the signature *)
           let sg =
             match fid with
-            | FunId (FRegular fid) | TraitMethod (_, _, fid) ->
+            | FunId (FRegular fid) ->
                 let trans_fun =
                   silent_unwrap __FILE__ __LINE__ span
                     (LlbcAst.FunDeclId.Map.find_opt fid trans_funs)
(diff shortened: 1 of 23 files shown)


Nadrieril:
Ah, this in fact doesn't pass typechecking in lean. I wasn't sure these methods were used. I guess I have to finish https://github.com/AeneasVerif/charon/issues/180 first.

sonmarcho:
> Ah, this in fact doesn't pass typechecking in lean. I wasn't sure these methods were used. I guess I have to finish [AeneasVerif/charon#180](https://github.com/AeneasVerif/charon/issues/180) first.

What doesn't typecheck? I don't see any issue in the diffs or in CI.

Nadrieril:
> What doesn't typecheck? I don't see any issue in the diffs or in CI.

That's because I just fixed it :) I had messed up the self clauses and didn't think I could easily fix it but I found a way.

sonmarcho:
Also, about the code you generate: at some point I thought of making the default methods take as inputs exactly the list of functions they need rather than a whole instance of the trait. Removing the input self clause when it is not needed is already a nice improvement.

Nadrieril:
> at some point I thought of making the default methods take as inputs exactly the list of functions they need

Yes we discussed this together :p

src/pure/PureMicroPasses.ml
```diff
@@ -3635,6 +3635,20 @@ let add_type_annotations_to_fun_decl (trans_ctx : trans_ctx)
...
+                log#ldebug
+                  (lazy (__FUNCTION__ ^ ": method name: " ^ method_name));
+                if Option.is_some lp_id then
+                  craise __FILE__ __LINE__ span
+                    "Trying to get a loop subfunction from a method call";
+                let method_sig =
+                  silent_unwrap __FILE__ __LINE__ span
+                    (Charon.Substitute.lookup_flat_method_sig trans_ctx.crate
+                       tref.trait_decl_ref.trait_decl_id method_name)
+                in
```
sonmarcho, line 3648: Could you add a TODO here? This works, but there should be a way of not invoking the translation there (by looking up the *pure* trait declaration)

sonmarcho: > We'd have to precompute these method signatures and store them somewhere

Yes, that's exactly what I have in mind: we would update the pure trait_decl definitino.

Issue #440, opened 2025-02-13
Lean type error from Rust translation


ayhon:
Consider the following minimized code snippet.
```rust
enum PeanoNum {
    Zero,
    Succ(Box<PeanoNum>),
}

fn minimized_bug(this: &mut PeanoNum, value: isize) {
    match this {
        PeanoNum::Zero => {
            std::mem::replace(this, PeanoNum::Succ(Box::new(PeanoNum::Zero)));
        }
        PeanoNum::Succ{..} => { }
    }
}
```
When translating it to Aeneas, I get the following
```lean
/- [testing_aeneas::PeanoNum]
   Source: 'src/lib.rs', lines 105:0-108:1 -/
inductive PeanoNum where
| Zero : PeanoNum
| Succ : PeanoNum → PeanoNum

/- [testing_aeneas::minimized_bug]:
   Source: 'src/lib.rs', lines 110:0-117:1 -/
def minimized_bug (this : PeanoNum) (value : Isize) : Result PeanoNum :=
  match this with
  | PeanoNum.Zero =>
    let (_, this1) :=
      core.mem.replace (Result.ok PeanoNum.Zero) (PeanoNum.Succ PeanoNum.Zero) -- TYPE ERROR (expects Result PeanoNum, receives PeanoNum)
    Result.ok this1 -- TYPE ERROR (expects PeanoNum, receives Result PeanoNum)
  | PeanoNum.Succ _ => Result.ok this
```

While minimizing my actual code into the previous sample, I notice the following:
- Adding a `bool` field to the `Zero` constructor seems to fix the translation. However, adding a `()` does nothing.
- If we add a new `PeanoNum::One` constructor to the `enum`, we can reproduce the error on a different branch if we use `PeanoNum.Succ PeanoNum.One` in the `std::mem::replace`. However, the error cannot be reproduced if these constructors reference one another. 

To clarify on the latter point, consider the following
```rust
enum PeanoNum {
    Zero,
    One,
    Succ(Box<PeanoNum>),
}

fn minimized_bug(this: &mut PeanoNum, value: isize) {
    match this {
        PeanoNum::Zero => {
            std::mem::replace(this, PeanoNum::Succ(Box::new(PeanoNum::One)));
        }
        PeanoNum::One => {
            std::mem::replace(this, PeanoNum::Succ(Box::new(PeanoNum::Zero)));
        }
        PeanoNum::Succ{..} => { }
    }
}
```
Running aeneas gives me the following
```lean
[...]

PR #441, opened 2025-02-13
Fix a micro-pass and update the names generated for the default methods

sonmarcho:
This PR fixes https://github.com/AeneasVerif/aeneas/issues/440 by fixing a bug in `PureMicroPasses.simplify_duplicate_calls`.
It also updates the names generated for the default methods, by appending the "default" keyword as a suffix (i.e., we now generate `BoolTrait.ret_true.default` instead of `BoolTrait.default.ret_true`).


tests/coq/rename_attribute/RenameAttribute.v
```diff
@@ -100,9 +100,9 @@ Definition no_borrows_sum (n : nat) (max : u32) : result u32 :=
   no_borrows_sum_loop n max 0%u32 0%u32
 .
 
-(** [rename_attribute::BoolTrait::default::ret_true]:
+(** [rename_attribute::BoolTrait::ret_true::default]:
     Source: 'tests/src/rename_attribute.rs', lines 15:4-17:5 *)
-Definition boolTrait_default_retTest
+Definition boolTrait_ret_true_retTest
```
Nadrieril, line 105: This should be `boolTrait_retTest_default`.

sonmarcho: Indeed. I think changing the *Rust* name is actually not a good idea. We should work directly on the extracted name.

Nadrieril: Or we could apply the renaming before this pass

Nadrieril: After private discussion, I see why changing the Rust name isn't great, objection witdrawn.

PR #451, opened 2025-03-11
feat(lean): tactic `progress*`

ayhon:
### Motivation
The purpose of the `progress` tactic is to ease the burden of the proof engineer when verifying the implementation of a definition generated by Aeneas. In particular, it allows one to more easily handle monadic code, through the use of a special type of "spec" theorems with a particular shape reminiscent of Hoare triples.
```lean
theorem spec
  (c: Rust.t)
: Precondition c
→ ∃ z, Rust.op c = .ok z ∧
  Postcondition c z
:= sorry
```

The `progress` tactic can be viewed as an analogous to the ["step"](https://sourceware.org/gdb/current/onlinedocs/gdb.pdf#84) operation in debuggers. Continuing with this analogy, `progress*` is meant to be an analogous to the ["finish"](https://sourceware.org/gdb/current/onlinedocs/gdb.pdf#85) operation in `gdb`, which continues running the program until then end of the function. Since the execution of a program through `progress` is symbolic, `progress*` needs to be able to follow all branches of a `match` or if-statement.

Sometimes `progress` is able to find a theorem to apply at a given step, but it's unable to prove some of its preconditions. When `progress*` is given a precondition tactic `tac` with the syntax `progress* by tac` it uses it to close goals generated by preconditions of `progress` theorems. 

## Objectives
1. Automate the repeated usage of `progress`.
2. Suggest the users an unfolding of `progress*` which shows the structure needed for the proof to conclude.  
```lean
def Rust.program(x: U8): Rust U8 := do
  let aux₁ <- Rust.add 1#u8 2#u8
  let aux₂ <- Rust.max 1#u8 aux₁
  let res <- Rust.add aux₂ aux₂
  if res < x then
    return res
  else
    let res₂ <- Rust.add res x
    return res₂

theorem Rust.program.spec (x: U8)
: x > 10#u8
→ ∃ z, program x = .ok z ∧ z.val = 6
:= by
  intro x_big
  unfold Rust.program
  -- This is suggested by using `progress*?`
  progress with Testing.Rust.add.spec as ⟨ aux₁, aux₁_post ⟩
  progress with Testing.Rust.max.spec as ⟨ aux₂, aux₂_post_1, aux₂_post_2, aux₂_post_3, aux₂_post_4 ⟩
  progress with Testing.Rust.add.spec as ⟨ res, res_post ⟩
  split
  case' intro.intro.isTrue _ =>
    skip
  case' intro.intro.isFalse _ =>
    progress with Testing.Rust.add.spec as ⟨ res₂, res₂_post ⟩
  -- Now one can conclude the goal
  all_goals (simp [pure]; scalar_tac)
```


backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
+import Lean
+import Aeneas.Utils
+import Aeneas.Progress.Progress
+import Aeneas.Progress.Core
+import Aesop.Util.Basic
+/- import Mathlib.Control.Sequence -/
```
sonmarcho, line 6: Not needed comment :)

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+open Lean Meta Elab Tactic 
+
+namespace Bifurcation/- {{{ -/
+/-- Expression on which a branch depends -/
+structure Discr where
+  toExpr: Expr
+  /-- Name under which to bind the discriminant equation, if provided -/
+  name?: Option Name := none
+  deriving Repr
+instance: ToMessageData Discr where
```
sonmarcho, line 17: I like having spaces between the various definitions, even though they are related

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+    m!"(info {kind} {discr} {branches})"
+
+def Info.ofExpr(e: Expr): MetaM (Option Info) := do
+  let e := e.consumeMData
+  if e.isIte || e.isDIte then
+    let kind := if e.isIte then .ite else .dite
+    --   ite.{u} {α : Sort u} (c : Prop) [h : Decidable c] (t           e : α)      : α
+    --  dite.{u} {α : Sort u} (c : Prop) [h : Decidable c] (t : c → α) (e : ¬c → α) : α
+    /- let #[_ty, cond, _dec, brThen, brElse] := e.getAppArgs -/
+    /- let e ← whnf e -/
```
sonmarcho, line 88: We should do a pass to clean this kind of comments :)

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+  if e.isIte || e.isDIte then
+    let kind := if e.isIte then .ite else .dite
+    --   ite.{u} {α : Sort u} (c : Prop) [h : Decidable c] (t           e : α)      : α
+    --  dite.{u} {α : Sort u} (c : Prop) [h : Decidable c] (t : c → α) (e : ¬c → α) : α
+    /- let #[_ty, cond, _dec, brThen, brElse] := e.getAppArgs -/
+    /- let e ← whnf e -/ 
+    let e ← deltaExpand e (fun n => n == ``ite || n == ``dite)
+    -- Decidable.casesOn.{u} {prop} {motive} dec (isFalse: (h:¬p) → motive (isFalse h)) (isTrue: (h:p) → motive (isTrue h)) : motive t
+    let .const ``Decidable.casesOn uLevels := e.getAppFn
+      | throwError "Expected ``Decidable.rec, found {←ppExpr e.getAppFn}"
```
sonmarcho, line 92: Is the error message correct (`Decidable.rec` vs `Decidable.casesOn`)?

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+    let #[prop, motive, dec, brFalse, brTrue] := e.getAppArgs
+      | throwError "Wrong number of parameters for {e.getAppFn}: {e.getAppArgs.size} [{e.getAppArgs}]"
+    return some {
+      kind,
+      discrs := #[{ toExpr := dec }]
+      -- TODO: I should be able to retrieve the name given to
+      --  the condition of a dite.
```
sonmarcho, line 99: I think the way to do this is to destruct the lambda of one of the branches and get the name given to the bound variable

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+    script := inf1.script ++ inf2.script,
+    unsolvedPreconditions := inf1.unsolvedPreconditions ++ inf2.unsolvedPreconditions,
+  }
+
+partial
+def evalProgressStar(cfg: Config): TacticM Info := withMainContext do
+  trace[ProgressStar] s!"Normalizing bind application in goal {←(getMainTarget >>= (liftM ∘ ppExpr))}"
+  Utils.simpAt (simpOnly := true) (thms := [``Aeneas.Std.bind_assoc_eq]) 
+    (loc := .targets #[] (type := true) )
+    (config := {}) (simprocs := []) (simpThms := [])
```
sonmarcho, line 183:
```suggestion
    (config := {maxDischargeSteps := 0}) (simprocs := []) (simpThms := [])
```
(just in case: it is good hygiene)

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+  trace[ProgressStar] s!"After bind normalization: {←ppExpr goalTy}"
+  let res ← Progress.programTelescope goalTy fun _xs _zs program _res _post => do
+    trace[ProgressStar] s!"Traversing {← ppExpr program}"
+    let resultName := .str .anonymous "res"
+    traverseProgram (onResult resultName) onBind onBif program
+  setGoals (res.unsolvedPreconditions ++ (←getGoals))
+  return res
+
+where
+  onResult resultName expr := do
```
sonmarcho, line 196: I don't fully understand this case: could you add a more detailed comment? (with maybe an example)

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,326 @@
...
+      return info
+
+  tryProgress := do
+    try some <$> Progress.evalProgress none none #[]
+    catch _ => pure none
+
+  trySolvePreconditions: TacticM (Array Syntax.Tactic × Array MVarId) := do
+    -- NOTE: Do I make the assumption that the preconditions appear before the final lemma, or
+    -- that the names of the preconditions contain the final lemma as a prefix.
+    -- For now, we have ←getUnsolvedGoals = preconditions ++ [final]
```
sonmarcho, line 261: I think we should update `evalProgress` so that it has type: `Array MVarId × Option MVarId` (list of preconditions + remaining goal, if there is still one)

sonmarcho (changes requested):
This is only a partial review (I have to jump into a meeting).
Overall, put aside minor details, everything I've seen looks good :)

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,346 @@
...
+  onResult resultName expr := do
+    trace[ProgressStar] s!"onResult: Since (· >>= pure) = id, we treat this result as a bind on id"
+    -- If we encounter `(do f a)` we process it as if it were `(do let res ← f a; return res)`
+    -- since (id = (· >>= pure)) and when we desugar the do block we have that 
+    -- 
+    --                      (do f a) == f a
+    --                               == (f a) >>= pure
+    --                               == (do let res ← f a; return res)
+    -- 
+    -- We known in advance the result of processing `return res`, which is to do nothing.
```
sonmarcho, line 217:
```suggestion
    -- We know in advance the result of processing `return res`, which is to do nothing.
```

backends/lean/Aeneas/Progress/ProgressStar.lean
```diff
@@ -0,0 +1,346 @@
...
+      | curr :: rest => do
+        setGoals [curr]
+        try
+          if let .some tac := cfg.preconditionTac then
+            evalTactic tac
+            return ←loop (acc.push <| ←`(tactic| · $(#[tac])*)) unsolved rest
+        catch _ => pure ()
+        let defaultTac ← `(tactic| · skip)
+        return ←loop (acc.push defaultTac) (unsolved.push curr) rest
+    loop #[] #[] (← getUnsolvedGoals)
```
sonmarcho, line 296: We should check whether the user provided a tactic *before* recursively going through the preconditions

sonmarcho (changes requested):

backends/lean/Aeneas/Progress/ProgressStar.lean
ayhon: Actually, this function is also in charge of removing the preconditions from the goal and constructing the script which handles them (which in that case would just be full of `· skip`). Perhaps we should rename the function to make this distinction

ayhon: I think `handleProgressPreconditions` is a more fitting name

sonmarcho: Oh indeed, I hadn't noticed it does all this in one go

ayhon: Yeah, it was not a great name :/

sonmarcho:
> A remark about [`traverseProgram`](https://github.com/ayhon/aeneas/blob/fele/backends/lean/Aeneas/Progress/ProgressStar.lean#L147-L173) and its use in [`evalProgressStar`](https://github.com/ayhon/aeneas/blob/ba92b57df2dbdf56985634320ddca770990c9718/backends/lean/Aeneas/Progress/ProgressStar.lean#L190-L205): `evalProgressStar` performs modifications on the TacticM state, dispatching tactics to solve goals while generating the suggestion script. However, `traverseProgram` itself does not make use of the tactic state, only looking at the given expression and traversing it. Therefore, it could happen that a discrepancy appears between what `evalProgressStar` believes the state of the goal should be and what the actual state of the goal is. In practice I believe this to be unlikely to happen, but it should be noted that it could be a possible future concern

I modified this to merge `traverseProgram` with `evalProgressStar`

PR #495, opened 2025-04-10
feat: add extra BitVec lemmas

ayhon:
Adds some convenience methods to `BitVec` to make it closer to `Vector`, and theorems relating to it.

In particular, we highlight
- `BitVec.set` to change bits of the `BitVec`
- An extensionality lemma for `BitVec`s


sonmarcho:
I made a wrong manip, so closing this PR which is subsumed by https://github.com/AeneasVerif/aeneas/pull/499

PR #502, opened 2025-04-16
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/655.

cc @sonmarcho to notify you that there's now `StorageLive`, `StorageDead` and `Deinit` in llbc. I merged `StorageDead` and `Deinit` into `Drop` as a pre-pass so nothing much has to change.


src/interp/InterpreterStatements.ml
```diff
@@ -866,9 +866,6 @@ and eval_statement_raw (config : config) (st : statement) : stl_cm_fun =
           let cc = cc_comp cc cf_assign in
           (* Compose and apply *)
           ([ (ctx, res) ], cc_singleton __FILE__ __LINE__ st.span cc))
-  | FakeRead p ->
-      let ctx, cc = eval_fake_read config st.span p ctx in
```
sonmarcho, line 870: Note to myself: the `eval_fake_read` function is not useful anymore

PR #504, opened 2025-04-21
Improve the Lean backend

sonmarcho:
This PR does the following:
 - introduce many lemmas about `List`, `Array`, `Vector` and `BitVec`
 - introduce a new tactic `simp_scalar` to simplify arithmetic expressions
 - improve `zmodify` to use `scalar_tac` as a discharger for the conditional rewritings
 - update Lean to 4.19.0-rc3
 - introduce the `simp_bool_prop_simps` simpset
 - improve progress to prioritize preconditions with a high number of meta-variables when looking up for assumptions with which to instantiate meta-variables, and to instantiate meta-variables before solving any precondition
 - improve and simplify `scalar_tac`
 - update `UScalar.sub_spec` so that the postcondition is: `exists res, ... /\ res.val = x.val - y.val` rather than `x.val = res.val + y.val`
 - add the example `mod_add` to the demo
 - add the `reduceNatEq` simproc to convert equalities of the shape `3 = e` to `e = 3` (the constant is on the right)
 - introduce the `olet` ("opaque let") and `tlet` ("transparent let") tactics to conveniently introduce let-bindings when doing proofs (those tactics automatically refold terms in the context)


sonmarcho:
This PR fixes https://github.com/AeneasVerif/aeneas/issues/513

PR #517, opened 2025-05-06
fix: core.slice.index.SliceIndexRangeFromUsizeSlice.index_mut

ayhon:
Previously #515, closed by mistake.


sonmarcho:
I updated the definition (again) and proved a sanity check lemma. Doing so forced me to introduce a few lemmas to reason about `getElem?`.

PR #526, opened 2025-05-22
feat: erase macro scopes for progress* generated identifiers

ayhon:
When `progress*` encountered a function which returned a product that was destructured in its progress theorem, some hygenic names would be used as part of the identifier. This change removes the macro scopes from those names.

### DIFF
--- a/backends/lean/Aeneas/Progress/ProgressStar.lean
+++ b/backends/lean/Aeneas/Progress/ProgressStar.lean
@@ -233,7 +233,7 @@ where
       else
         trace[ProgressStar] "onBind: all preconditions solved"
       /- Update the main goal, if necessary -/
-      let ids ← getIdsFromUsedTheorem name usedTheorem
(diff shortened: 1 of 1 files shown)


sonmarcho:
Nice. Do you have an example where that makes a difference?

ayhon:
The hygenic name comes from the translation `let (x, y) ← foo; const x y` to `let h ← foo; match h with | (x,y) => cont x y`, which we simplify using projections when executing the following `simp` in `progress*`.
https://github.com/AeneasVerif/aeneas/blob/8062040edab6e9f725ef68cf04d4288acef0f6f5/backends/lean/Aeneas/Progress/ProgressStar.lean#L155-L158

If I remember correctly, we had this `simp` because: (1) we wanted to normalize the order of `>>=` so they were all left-associative and (2) to prevent having matches of only one constructor (which is an assumption `onBif` makes).

PR #535, opened 2025-06-05
Register charon into the current opam switch after rebuilding it

Remyjck:
This should be non-destructive as `opam install` succeeds even if the library already exists.

### DIFF
--- a/scripts/check-charon-install.sh
+++ b/scripts/check-charon-install.sh
@@ -8,9 +8,9 @@ fi
 
 rebuild() {
     if which nix 2> /dev/null 1>&2; then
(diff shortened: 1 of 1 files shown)


Remyjck:
Ah that's possible. 
I installed Aeneas without being a charon developer (so without using a symlink), and in my case I needed to manually install it to my opam switch.

Remyjck:
I've looked into it a little further.

It seems that my issue was tied to the fact that I was using a WSL with files stored on the Windows OS.
I've moved the project into my Linux filesystem and opam is now able to read the symlink.

My situation should be rare, so it should actually be fine to keep things as they are.

PR #538, opened 2025-06-10
Update charon

Nadrieril:
This causes a large diff because of definitions changing order. We could keep `all_ids` in charon and just make it more robust to forgetting to add ids in it, but that's convoluted.

cc @sonmarcho is this diff acceptable?


tests/lean/Demo/Demo.lean
```diff
@@ -8,6 +8,31 @@ set_option linter.unusedVariables false
...
+  let self1 ← self + 1#usize
+  ok (self, self1)
+
+/- Trait implementation: [demo::{demo::Counter for usize}]
+   Source: 'tests/src/demo.rs', lines 104:0-110:1 -/
+@[reducible]
+def CounterUsize : Counter Usize := {
+  incr := CounterUsize.incr
+}
+
```
sonmarcho, line 35: This is a problematic reordering: no definition in the `Demo.lean` file should move.
As a side note: when I give demos during presentations I show the Rust file and the Lean file side by side. If the order was not the same, the tool would look a bit broken.

Nadrieril:
@sonmarcho Items are now sorted by crate, file name, and span. This should make the output a lot easier to follow for users. It that diff ok?

sonmarcho (commented):
That's a lot better, thanks! But I see there are still a few (surprising) issues. There is probably a bug in the reordering?

sonmarcho:
I fixed the issue. Here is a minimal example of what happened:
```rust
pub struct Before();

#[derive(Clone)]
pub struct After();
```
In the extracted code, `After` appears before `Before`. The reason is that because the default `hashmap::{impl Clone for After}::clone_from` method generated for `After` receives the span from the *original* default implementation (from the standard library), this implementation is put at the beginning of the item list (in particular, before `Before`). As `hashmap::{impl Clone for After}::clone_from` depends on `After`, it means that `After` is moved before this item (i.e., before `Before`).

By modifying the reordering of Charon so that duplicated default methods use the span of the implementation rather than the span of the original default method we get what we expect.

sonmarcho (approved):

PR #564, opened 2025-07-03
FCongr, tlet modification, simp_scalar modification

JOSHCLUNE:
I have confirmed that these changes do not break any of the tests in aeneas/tests/lean or any of the proofs in the josh/proofs branch of symcrust

### DIFF
--- a/backends/lean/Aeneas/FCongr/FCongr.lean
+++ b/backends/lean/Aeneas/FCongr/FCongr.lean
@@ -1,0 +1,18 @@
+import Lean
+
+open Lean Meta Parser Elab Tactic
+
+namespace FCongr
+
+syntax (name := fcongr) "fcongr" (ppSpace num)? : tactic
+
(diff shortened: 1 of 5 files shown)


backends/lean/Aeneas/FCongr/FCongr.lean
```diff
@@ -0,0 +1,18 @@
...
+namespace FCongr
+
+syntax (name := fcongr) "fcongr" (ppSpace num)? : tactic
+
+@[tactic fcongr]
+def evalFCongr : Tactic
+| `(tactic| fcongr $[$n?]?) => withMainContext do
+  let hugeDepth := 1000000
+  let depth := n?.map (·.getNat) |>.getD hugeDepth
+  withTransparency .reducible do
```
sonmarcho, line 14: I guess changing the transparency is the trick which makes it work? What is the default reducibility used by `congr`?

PR #579, opened 2025-07-19
Update Charon

firefighterduck:
Companion to https://github.com/AeneasVerif/charon/pull/771

### DIFF
--- a/src/extract/Extract.ml
+++ b/src/extract/Extract.ml
@@ -378,7 +378,13 @@ let extract_unop (span : Meta.span) (extract_expr : bool -> texpression -> unit)
                    (PrintPure.integer_type_to_string ty)
              in
              match (src, tgt) with
-             | TInteger src, TInteger tgt ->
+             | _, _
+               when ValuesUtils.literal_type_is_integer src
+                    && ValuesUtils.literal_type_is_integer tgt ->
+                 let src, tgt =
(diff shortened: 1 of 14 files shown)


src/extract/Extract.ml
```diff
@@ -378,7 +378,13 @@ let extract_unop (span : Meta.span) (extract_expr : bool -> texpression -> unit)
                    (PrintPure.integer_type_to_string ty)
              in
              match (src, tgt) with
-             | TInteger src, TInteger tgt ->
+             | _, _
+               when ValuesUtils.literal_type_is_integer src
+                    && ValuesUtils.literal_type_is_integer tgt ->
```
sonmarcho, line 383: I didn't have time to review my notifications on Friday and just noticed this.
This way of updating the pattern matching is quite strange and comes from the fact that `LiteralTy` now has the `Int` and `UInt` variants. Why not split `IntegerTy` between `Signed` and `Unsigned`, **but keep** the `Integer` variant in `LiteralTy`? In particular I note that `IntegerTy` seems to not be used anymore in the AST, and to only be a convenience definition (for situations like this one). Does removing the `Integer` variant make that much difference elsewhere?

Nadrieril: I'm conflicted on this, because `bool` and `char` are also sometimes sufficiently integer-like that grouping them with integers could make sense.

PR #590, opened 2025-08-05
Escherichia/match on char

EschericHya:
Support for match on char for charon https://github.com/AeneasVerif/charon/issues/546

I'm not sure if I should add more support or if passing the tests is enough

### DIFF
--- a/charon-pin
+++ b/charon-pin
@@ -1,2 +1,2 @@
 # This is the commit from https://github.com/AeneasVerif/charon that should be used with this version of aeneas.
-0bc8c18a8dea17754abba0eeffd3204d996aff9e
+cee6178c7003efe014bba333c3932c18bddd2ec3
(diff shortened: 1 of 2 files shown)


src/interp/InterpreterStatements.ml
```diff
@@ -983,20 +983,24 @@ and eval_switch (config : config) (span : Meta.span) (switch : switch) :
...
-  | SwitchInt (op, int_ty, stgts, otherwise) ->
+  | SwitchInt (op, (int_ty : literal_type), stgts, otherwise) ->
       (* Evaluate the operand *)
       let op_v, ctx, cf_eval_op = eval_operand config span op ctx in
       (* Switch on the value *)
       let ctx_resl, cf_switch =
-        match op_v.value with
-        | VLiteral (VScalar sv) -> (
+        match (op_v.value, int_ty) with
+        | VLiteral (VScalar sv), TInt _ | VLiteral (VScalar sv), TUInt _ -> (
```
Nadrieril, line 992: I think that's legal:
```suggestion
        | VLiteral (VScalar sv), (TInt _ | TUInt _) -> (
```

Issue #500, opened 2025-04-13
Incorrect extraction on loop with copy


ayhon:
MWE:
```rust
fn bar(a: &mut bool) { }
fn foo(s: &mut bool) {
    let mut a = *s;
    while 0 < 0 {
        bar(&mut a);
    }
    *s = a;
}
```

This code is extracted to
```lean4
-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
-- [foo]
import Aeneas
open Aeneas.Std Result Error
set_option linter.dupNamespace false
set_option linter.hashCommand false
set_option linter.unusedVariables false

namespace foo

/- [foo::bar]:
   Source: 'src/foo.rs', lines 1:0-1:24 -/
def bar (a : Bool) : Result Bool :=
  ok a

/- [foo::foo]: loop 0:
   Source: 'src/foo.rs', lines 4:4-6:5 -/
def foo_loop (s : Bool) : Result Unit :=
  if 0#i32 < 0#i32
  then do
       let a ← bar s
       foo_loop a
  else ok ()
partial_fixpoint

/- [foo::foo]:
   Source: 'src/foo.rs', lines 2:0-8:1 -/
def foo (s : Bool) : Result Bool :=
  do
  foo_loop s
  ok s

end foo
```

In the extracted code, `foo_loop` is a no-op, while it should be returning `Bool` since it takes a mutable reference to `s` (regardless of whether `foo_loop` performs any changes to `s` or not).

ayhon:
While minimizing, I encountered that this code
```rust
struct A([bool; 1]);

fn bar(a: &mut [bool; 1]) {}

fn foo(s: &mut [bool; 1]) {
    let mut a = A(*s);
    while false {
        bar(&mut a.0);
    }
    *s = a.0;
}
```
gives the following error
```
[Info ] Imported: foo.llbc
[Info ] Generated the partial file (because of errors): ./Foo.lean
[Error] Can't end abstraction 0 as it is set as non-endable
Source: 'src/foo.rs', lines 7:4-9:5
```

ayhon:
Also, the following code
```rust
struct A([bool; 1]);

fn foo(s: &mut [bool; 1]) {
    let mut a = A(*s);
    while 0 < 0 { }
    *s = a.0;
}
```
results in the following crash
```
[Info ] Imported: foo.llbc
Uncaught exception:

  "Assert_failure charon-ml/src/Collections.ml:207:4"

[...]

Issue #355, opened 2024-11-12
Fix `InterpreterPaths.access_projection`


sonmarcho:
The `access_projection` function does not exactly implement the semantics of LLBC when dereferencing a shared borrow.
At [this line](https://github.com/AeneasVerif/aeneas/blob/2ad24ac2ec6eefc425250068d7caebba360871f2/compiler/InterpreterPaths.ml#L176) we should dive inside the shared value **only** if the remaining path is not empty.

Note that the impact of this semantical change is very minor, so it's not a high priority.

Issue #351, opened 2024-11-10
Bug in loop fixed-point computation


sonmarcho:
This triggers a bug (comes from investigating https://github.com/AeneasVerif/aeneas/issues/270 - it is a different bug):
```rust
pub enum List<T> {
    Cons(T, Box<List<T>>),
    Nil,
}

fn foo(h : &u8, mut t: &List<u8>) {
    let mut last = h;
    while let List::Cons(ht, tt) = t
    {
        last = ht;
        t = &*tt;
    }
}
```

The error is:
```
[Error] In file InterpreterLoopsFixedPoint.ml, line 680:
The sets of abstractions we need to end per region group are not pairwise disjoint
```

If we use the same lifetime for the two shared borrows, the translation succeeds:
```rust
pub enum List<T> {
    Cons(T, Box<List<T>>),
    Nil,
}

fn foo<'a>(h : &'a u8, mut t: &'a List<u8>) {
    let mut last = h;
    while let List::Cons(ht, tt) = t
    {
        last = ht;
        t = &*tt;
    }
}
```

The issue will be solved once we finish generalizing the support for loops (to add nested loops for instance).

Issue #270, opened 2024-06-27
Unreachable List Loop


Thanhson89:
```rust
use std::borrow::Borrow;

pub enum List<T> {
    Cons(T, Box<List<T>>),
    Nil,
}

fn foo(v: &List<List<u8>>) -> Option<List<u8>> {
    if let List::Cons(h, t) = v {
        let mut t = t.borrow();
        let mut last = h;
        while let List::Cons(ht, tt) = t
        {
            last = ht;
            t = tt.borrow();
        }
        None
    }
    else {
        None
    }
}
```

Trigger

$ aeneas -backend lean list_cons.llbc 
[Info ] Imported: list_cons.llbc
[Info ] Generated the partial file (because of errors): ./ListCons.lean
[Error] Unreachable
Source: 'src/main.rs', lines 11:8-18:5

sonmarcho:
Minimized further:
```rust
pub enum List<T> {
    Cons(T, Box<List<T>>),
    Nil,
}


fn foo(v: &List<List<u8>>) {
    if let List::Cons(_, t) = v {
        let mut t = &**t;
        while let List::Cons(_, tt) = t
        {
            t = &**tt;
        }
    }
}
```

When using the option `-abort-on-error` the log is:
```
[Info ] Imported: rust_tests.llbc
[Error] In file InterpreterLoopsFixedPoint.ml, line 531:
Unreachable
Source: 'src/lib.rs', lines 10:8-13:9

Uncaught exception:
  
  (Failure
    "In file InterpreterLoopsFixedPoint.ml, line 531:\
   \nIn file InterpreterLoopsFixedPoint.ml, line 531:\
   \nUnreachable\
   \nSource: 'src/lib.rs', lines 10:8-13:9\
   \nSource: 'src/lib.rs', lines 10:8-13:9")

Raised at Aeneas__Errors.craise_opt_span in file "Errors.ml", line 52, characters 4-76
Called from Stdlib__List.filter_map.aux in file "list.ml", line 258, characters 14-17
Called from Aeneas__InterpreterLoopsFixedPoint.compute_loop_entry_fixed_point.compute_fixed_ in file "InterpreterLoopsFixedPoint.ml", line 542, characters 26-68
Called from Aeneas__InterpreterLoopsFixedPoint.compute_loop_entry_fixed_point in file "InterpreterLoopsFixedPoint.ml", line 574, characters 11-60
[...]

PR #600, opened 2025-08-22
Make the joins and the translation of loops more general

sonmarcho:
The computation of loops fixed points and in particular the joins of contexts hadn't evolved much since its first version, which was mostly a prototype. This PR performs a complete overall of this part of the symbolic execution and the translation, fixing many issues in the process.

One of the key elements was to introduce computational content to region abstractions, which take as inputs the values consumed upon ending loans and output the values given back upon ending borrows. The consequence is that region abstractions can now represent arbitrary continuations, and not simply backward functions introduced by function calls or loops. The benefit is that when joining two contexts, we do not have to end loans which appear on one side and not the other. Moreover, this allowed lifting constraints about loop region abstractions, enabling us to support tricky use cases like the following one:

```rust
fn issue400_2(a: &mut i32, b: &mut i32, c: &mut i32, conds : &[bool]) {
    let mut y = &mut *a;
    let mut z = &mut *b;
    let mut i = 0;
    while i < conds.len() {
        // y and z borrow a, b, and c, and we don't know exactly how
        // (i.e., there are region abstractions linking the borrows in y and z
        // and the loans in a, b and c)
        if conds[i] {
            // we have to end the borrows in y and z before introducing new borrows
            y = &mut *a;
            z = &mut *b;
        }
        else {
            // we have to end the borrows in y and z before introducing new borrows
            y = &mut *b;
            z = &mut *c;
        }
        // y and z borrow either a and b, or b and c
        i += 1;
    }
    // y and z borrow either a and b, or b and c
    *y += 3;
    *z += 5;
}
```

which gets translated to:
```lean
/- The loop receives a backward function as input, which represents the computational
    content of the region abstraction which links y and z to a, b and c -/
def issue400_2_loop
  (back : I32 → I32 → (I32 × I32 × I32)) (conds : Slice Bool) (y : I32)
  (z : I32) (i : Usize) :
  Result (I32 × I32 × (I32 → I32 → (I32 × I32 × I32)))
  :=
  let i1 := Slice.len conds
  if i < i1
  then
    do
    let b ← Slice.index_usize conds i
    if b
    then
      do
      let (a, b1, c) := back y z
      let i2 ← i + 1#usize
      issue400_2_loop (fun i3 i4 => (i3, i4, c)) conds a b1 i2
    else
      do
      let (a, b1, c) := back y z
      let i2 ← i + 1#usize
      issue400_2_loop (fun i3 i4 => (a, i3, i4)) conds b1 c i2
  else ok (y, z, back)
partial_fixpoint

def issue400_2
  (a : I32) (b : I32) (c : I32) (conds : Slice Bool) :
  Result (I32 × I32 × I32)
  :=
  do
  let (y, z, back) ←
    issue400_2_loop (fun i i1 => (i, i1, c)) conds a b 0#usize
  let y1 ← y + 3#i32
  let z1 ← z + 5#i32
  ok (back y1 z1)
```

Loops are now characterized by an entry fixed point but also by an exit context, which is the result of joining the contexts at the breaks and which is the context with which to continue the execution after the loop. This allows us to make the translation cleaner (what happens after the loop is no longer inlined inside the loop) and support nested loops (as long as they do not contain breaks/continues to outer loops or early returns) as well as functions containing sequences of loops. This PR also opens the possibility of not duplicating the control-flow after a match or an if then else (this will be done in a subsequent PR).

We also performed the following changes:
- we removed the `-state` CLI option because it had become cumbersome to support and will not be useful at all once we introduce support for separation logic. This also required removing some tests and in particular the b-epsilon tree, which was also not really useful (it contained no proofs, and has never allowed us to catch any bug in the translation)
- we also temporarily disabled support for (infinite) loops which contain no breaks.

Fixes: https://github.com/AeneasVerif/aeneas/issues/355 (had to fix this to make some merges of environments work)
Fixes: https://github.com/AeneasVerif/aeneas/issues/401 (the pass was removed altogether, as it is now useless)
Fixes: https://github.com/AeneasVerif/aeneas/issues/500, https://github.com/AeneasVerif/aeneas/issues/351, https://github.com/AeneasVerif/aeneas/issues/270 (generalizing the loops fixes this issue)


src/pure/PureMicroPasses.ml
```diff
@@ -2836,202 +3195,2572 @@ let add_fuel_and_state_one (ctx : ctx) (loops : fun_decl LoopId.Map.t)
...
+       so it should be driven by examples. *)
+    let input_to_output =
+      ref
+        (Collections.IntMap.of_list
+           (List.init loop.num_input_conts (fun i -> (i, []))))
+    in
+    List.iter
+      (fun (output_index, input_index) ->
+        input_to_output :=
+          Collections.IntMap.add_to_list input_index output_index
```
oliver-butterley: When I try to `make` aeneas I get an error from this line:

Image

PR #604, opened 2025-09-03
chore: fix proof for forward compatibility with v4.22.0

kim-em:
`letFun_val_congr` was an internal implementation detail which disappears on `v4.22.0`. `bind_congr` is a better choice at this point.


sonmarcho:
Thanks for this!

PR #618, opened 2025-10-03
Removed some charon rename

EschericHya:
Removed some `charon::rename` as discussed with @Nadrieril


sonmarcho:
I would like to wait a bit before merging this: I'm wrapping up the PR about the loops, and there will be collisions.

PR #660, opened 2025-11-27
Update charon

N1ark:
Partner to https://github.com/AeneasVerif/charon/pull/920


src/interp/InterpExpressions.ml
```diff
@@ -450,7 +450,7 @@ let eval_operand_no_reorganize (config : config) (span : Meta.span)
                 e )
           in
           (cv, ctx, cc_comp cc cf)
-      | CFnPtr _ -> [%craise] span "Function pointers are not supported yet"
+      | CFnDef _ -> [%craise] span "Function definitions are not supported yet"
```
sonmarcho: This error message is a bit weird from a user point of view

PR #711, opened 2026-01-16
Start adding models for reasoning about array/vec iteration

R1kM:
Adds the required models in the Lean std to extract the small iterators-array test.
Also slightly extends the previous iterators-scalar test.

### DIFF
--- a/charon-pin
+++ b/charon-pin
@@ -1,2 +1,2 @@
 # This is the commit from https://github.com/AeneasVerif/charon that should be used with this version of aeneas.
-051a0322504d97dea97f013be0c70a6b48811bf0
+56e0268dbfae33c07b5f0721e8cbf304f42faf55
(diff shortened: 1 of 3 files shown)


backends/lean/Aeneas/Std/Core/Iter.lean
```diff
@@ -135,3 +135,21 @@ def core.iter.traits.iterator.IteratorRange {A : Type} (StepInst :
   (core.ops.range.Range A) A := {
   next := core.iter.range.IteratorRange.next StepInst
 }
+
+@[reducible, rust_trait_impl
+  "core::iter::traits::iterator::Iterator<alloc::vec::into_iter::IntoIter<@T, @A>, @T>"
+  (keepParams := [true, false])]
+def core.iter.traits.iterator.IteratorIntoIter (T : Type) :
```
sonmarcho: Note to self: rename to mention `vec`

PR #819, opened 2026-03-03
Bug fix: leadingZeros

fharding1:
The previous implementation of leadingZeros had a bug: it would return 0 for the 0^w bitvec, instead of w. This is not how Rust's leading_zeroes is defined.
```rust
fn main() {
    let a: u64 = 0;
    println!("{}", a.leading_zeros()); // prints 64
}
```

I have updated it here to fix the bug and also use Nat.log instead of a recursive definition using getMsbD. This might be more natural since BitVec uses Fin n under the hood rather than List Bool. It also makes it easier to write certain kinds of proofs which relate leading_zeroes to the size of the number (my use case). It could be nice to keep the recursive definition and have a theorem equating them, thoughts?


sonmarcho:
Thanks for this! I'm not that interested in keeping the old definition. What I care more about is the possibility of easily proving statements like the following one:
```lean
forall i < x.leadingZeros, x[w - i - 1] = 0
```

PR #832, opened 2026-03-11
Rename BitVec.toNat_pow BitVec.ofNat_pow to avoid conflict with Batte…

klausnat:
renaming 

BitVec.toNat_pow to Aeneas.BitVec.toNat_pow
BitVec.ofNat_pow to Aeneas.BitVec.ofNat_pow

to avoid conflict with mathlib/Batteries


oliver-butterley (commented):
It could make sense to use `namespace BitVec` in `BitVec.lean` so the lemmas just added don't need the prefix. Additionally it would mean that all the theorems definitions can be written without the prefix. Seems natural!

Great about this correction, we also had the same issue in our repo using Aeneas.

PR #837, opened 2026-03-12
feat: Tactic to infer post conditions

### LINKED ISSUE #826: Feat: implement a tactic to infer post-conditions

mpenciak:
This PR adds logic to instantiate postcondition metavariables that come up in applications of `step` and `step*` for higher order `step` lemmas. The logic is contained in three places:

1. An `infer_post` tactic that can be invoked when the goal state is of the form `?post args,*`
2. A configuration option for `step` so that `step +inferPost` will unify and close metavariable goals that may come up in the form `step _ {?post args,*}`
3. A configuration option for `step*` so that `step* +inferPost` can make progress for the same goals as above.

Closes #826


sonmarcho (commented):
That's great, thanks! I'm being very nitpicky so there is a ton of comments but they're all quite minor.

PR #885, opened 2026-03-27
chore: replace custom List.slice  with upstream List.extract

### LINKED ISSUE #814: Remove `List.slice` and use `List.extract` instead

oliver-butterley:
List.slice -> List.extract refactor

Replaced the custom List.slice definition with Lean's built-in List.extract (identical semantics, different arg order: slice start stop ls -> ls.extract start stop).

Changes:

- List.lean: Removed def slice, added a deprecated abbrev slice alias. Renamed 8 theorems (slice_nil → extract_nil, etc.) with deprecated aliases for the old names. Simplified proofs since extract unfolds via simp.
- Slice.lean: Slice.slice body updated to delegate to List.extract. Updated internal theorem references. 
- Array.lean: Array.slice body updated to delegate to List.extract.
- ArraySlice.lean, Vec.lean: Updated theorem statements from .val.slice to .val.extract.
- Slice.slice and Array.slice keep their names (This is the desired behaviour since it matches the Rust terminology?).

fixes #814


mpenciak (approved):
This looks good to me! ~There are a few renamed Slice lemmas that weren't aded to the deprecation list. I'm not sure if it's super important, but it's the only thing I noticed.~ (looking closer I actually didn't catch everything, everything looks good to me!)

backends/lean/Aeneas/Data/List/List.lean
```diff
@@ -220,43 +229,45 @@ theorem resize_length (l : List α) (new_len : Nat) (x : α) :
...
   (_ : j ≤ ls.length ∧ i + k < j) :
-  (ls.slice i j)[k]? = ls[i + k]? := by
-  revert i j
-  induction ls
-  . intro i j; simp_all
-  . intro i j h
-    simp_all [slice]
-    have : k < j - i := by scalar_tac
-    simp [*]
+  (ls.extract i j)[k]? = ls[i + k]? := by grind
```
sonmarcho, line 251: Do we actually still need this theorem (as `grind` solves the goal)?
In particular: I'm wondering whether we should mark it as `grind =`
It is very likely that we just need to mark a Lean std lemma with `simp_lists_safe`

backends/lean/Aeneas/Data/List/List.lean
```diff
@@ -220,43 +229,45 @@ theorem resize_length (l : List α) (new_len : Nat) (x : α) :
...
+alias getElem?_slice := getElem?_extract
 
 @[simp, simp_lists_safe, grind =]
-theorem getElem!_slice [Inhabited α] (i j k : Nat) (ls : List α)
+theorem getElem!_extract [Inhabited α] (i j k : Nat) (ls : List α)
   (_ : j ≤ ls.length ∧ i + k < j) :
-  (ls.slice i j)[k]! = ls[i + k]! := by
-  have := getElem?_slice i j k ls
-  simp_all
+  (ls.extract i j)[k]! = ls[i + k]! := by grind
```
sonmarcho, line 259: same remark as above

backends/lean/Aeneas/Data/List/List.lean
```diff
@@ -220,43 +229,45 @@ theorem resize_length (l : List α) (new_len : Nat) (x : α) :
...
-theorem getElem_slice (i j k : Nat) (ls : List α)
+theorem getElem_extract (i j k : Nat) (ls : List α)
   (_ : j ≤ ls.length ∧ i + k < j) :
-  (ls.slice i j)[k] = ls[i + k] := by
-  have hk : k < (ls.slice i j).length := by simp [slice_length]; omega
-  have hik : i + k < ls.length := by omega
-  have h1 := getElem?_slice i j k ls (by grind)
-  rw [getElem?_eq_getElem hk, getElem?_eq_getElem hik] at h1
-  exact Option.some.inj h1
+  (ls.extract i j)[k] = ls[i + k] := by grind
```
sonmarcho, line 267: Same remark

oliver-butterley:
Closing this since the base objects are now defined using `ListN` (https://github.com/AeneasVerif/aeneas/pull/1256) and so this is no longer relevant. @sonmarcho

Issue #649, opened 2025-11-25
Add models and progress theorems for scalar unops and binops


hatoo:
Currently, we can't use progress tactic on xor expression due to lack of spec for xor.

```lean
@[progress]
theorem UScalar.xor_spec {ty} (x y : UScalar ty) :
  ∃ z, toResult (x ^^^ y) = ok z ∧ z.val = (x ^^^ y).val ∧ z.bv = x.bv ^^^ y.bv := by
  simp [toResult]
  rfl
```

`UScalar.xor_spec` should be in the library.

PR #893, opened 2026-03-31
Add missing step theorems for scalar negation, signed shifts, and rotates

MavenRain:
Fixes rotate_right bug (was using rotateLeft), adds IScalar.neg_spec, IScalar shift specs, rotate specs, and IScalar val_* simp lemmas.

Addresses #649.

### DIFF
--- a/documentation/skills/aeneas-crypto-verification.instructions.md
+++ b/documentation/skills/aeneas-crypto-verification.instructions.md
@@ -76,11 +76,59 @@ vacuity test.
 
 ## Function Decomposition
 
-### When to use:
+### Phase analysis (mandatory — do BEFORE attempting any proof)
 
-- Function body has >10 monadic steps
-- Contains repeated sub-patterns (e.g., Montgomery reduction appears multiple times)
-- Proof is timing out or unmanageable
+Phase analysis and fold decomposition are **MANDATORY** before writing any proof
+for a function body with >10 monadic steps. Do NOT attempt a direct proof
+and then decompose after it fails — analyze and decompose first.
+
+**Analyze the body** and identify distinct **algorithmic phases**. An algorithmic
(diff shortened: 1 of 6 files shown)


backends/lean/Aeneas/Std/Scalar/Bitwise.lean
```diff
@@ -280,4 +328,9 @@ theorem IScalar.not_spec {ty} (x : IScalar ty) :
 @[simp, scalar_tac_simps, grind =, agrind =] theorem UScalar.val_or {ty} (x y : UScalar ty) : (x ||| y).val = x.val ||| y.val := by rfl
 @[simp, scalar_tac_simps, grind =, agrind =] theorem UScalar.val_xor {ty} (x y : UScalar ty) : (x ^^^ y).val = x.val ^^^ y.val := by rfl
 
+@[simp, scalar_tac_simps, grind =, agrind =] theorem IScalar.val_and {ty} (x y : IScalar ty) : (x &&& y).val = (x.bv &&& y.bv).toInt := by rfl
+@[simp, scalar_tac_simps, grind =, agrind =] theorem IScalar.val_or {ty} (x y : IScalar ty) : (x ||| y).val = (x.bv ||| y.bv).toInt := by rfl
+@[simp, scalar_tac_simps, grind =, agrind =] theorem IScalar.val_xor {ty} (x y : IScalar ty) : (x ^^^ y).val = (x.bv ^^^ y.bv).toInt := by rfl
+@[simp, scalar_tac_simps, grind =, agrind =] theorem IScalar.val_not {ty} (x : IScalar ty) : (~~~x).val = (~~~x.bv).toInt := by rfl
```
sonmarcho: I'm not sure this is what we want. I guess it is not possible to prove something like `(x &&& y).val = x.val &&& y.val`?

PR #918, opened 2026-04-07
feat: New `do` Elaborator

mpenciak:
We implement a new `do` elaborator for extracted monadic code. The new elaborator is only triggered for `do` blocks of type `Result`, and there is a new option `Aeneas.newDoElab` which can be unset to use the old elaborator.

The new elaborator works on a subset of the `do` syntax which doesn't include early returns, so we can avoid any do jumps and the resulting expression will be a sequence of binds.

We also treat the pattern matching syntax in `let` bindings using `Function.uncurry` to avoid the built-in `match` syntax elaboration.

Included is a new delaborator that delabs the new expressions to the expected syntax.


backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -0,0 +1,492 @@
...
+  fun _ k =>
+    Meta.lambdaTelescope arm fun fvars ebody => do
+      let .mvar mvarId := ebody.getAppFn
+        | throwError "elabDoMatch: expected metavariable arm body, got{indentExpr ebody}"
+      let armExpr ← (body.run ctx).run pure
+      let value ←
+        if ebody.isMVar then
+          let usedFVars := fvars.filter (armExpr.occurs ·)
+          if usedFVars.isEmpty then pure armExpr
+          else mkLambdaFVars usedFVars armExpr
```
sonmarcho, line 102: I don't fully understand why we need to filter the fvars

backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -0,0 +1,492 @@
...
+        else
+          mkLambdaFVars fvars armExpr
+      mvarId.assign value
+      k ()
+
+
+/-- For a single-constructor inductive (other than `Prod`), return the constructor
+    name and field types instantiated at the type's parameters. `none` otherwise. -/
```
sonmarcho, line 109: Why does `Prod` have its own treatment?

backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -0,0 +1,492 @@
...
+        discard <| isDefEq α a
+      pure e
+  let α ← instantiateMVars α
+  let shape ← analyzePat pat α
+  let k ← mkPatContinuation shape α fun _ => elabDoSeqCore rest
+  synthesizeSyntheticMVarsNoPostponing
+  ElabM.mkBind e k
+
+/-- Elaborate `let x := e`. -/
+partial def elabDoLetId (x : Ident) (ty? : Option Term) (rhs : Term)
```
sonmarcho, line 384: Isn't this a sub-case of `elabDoLetPat`?

backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -0,0 +1,492 @@
...
+      let expectedType ← ElabM.mkMonadicType unitType
+      let e ← elabAtType expectedType
+      ElabM.mkBind e body
+
+/-- Elaborate a term in a `doSeq` position. -/
+partial def elabDoExpr (term : Term) (rest : List DoElem) : ElabM Expr :=
+  elabMonadicAsDoElem (fun ty => elabTermEnsuringType term ty) rest
+
+/-- Elaborate `let x ← e`. -/
+partial def elabDoLetArrowId (x : Ident) (ty? : Option Term) (rhs : DoElem)
```
sonmarcho, line 333: Isn't this a sub-case of `elabDoLetArrowPat`?

backends/lean/Aeneas/Do/Delab.lean
```diff
@@ -0,0 +1,197 @@
...
+  let valStx ← withAppArg delab
+  let (pats, bodyStx) ← withAppFn <| withAppArg <|
+    enterUncurryChain #[] fun outerBinders => do
+      delabBinders outerBinders.toList delab
+  let tupleStx ← buildTupleTerm pats
+  `(let $tupleStx:term := $valStx
+    $bodyStx)
+
+/-- `T.casesOn val (fun f₁ … fₙ => body)` → `let ⟨f₁, …, fₙ⟩ := val; body` -/
+@[delab app]
```
sonmarcho, line 99: Can we make this `delab` annotation more precise (it seems we can't but it's worth investigating)? Also, it seems this is one linked to the new delaborator and triggers on all `casesOn` with a single branch, right (just confirming)?

sonmarcho (approved):
Looks good, thanks! I'm going to merge this as it is: we can address the comments in a subsequent PR.

I initially wanted to leave some comments in the code but they block the merge, so I'm listing them below instead.

- `ElabM.assignArmMVar`: I don't understand why we need to filter the fvars in the branch `if ebody.isMVar`
- `getCtorFieldTypes`: why does `Prod` have its own treatment?
- `elabDoLetId `: isn't this a sub-case of `elabDoLetPat`?
- `elabDoLetArrowId `: isn't this a sub-case of `elabDoLetArrowPat `?

PR #919, opened 2026-04-08
Bump Charon

N1ark:
Partner PR to https://github.com/AeneasVerif/charon/pull/1088

### DIFF
--- a/src/PrePasses.ml
+++ b/src/PrePasses.ml
@@ -62,7 +62,7 @@ let erase_body_regions (crate : crate) (f : fun_decl) : fun_decl =
   (* Map  *)
   let body =
     match f.body with
-    | Some body ->
+    | Body body ->
         let body =
           {
             body with
@@ -74,8 +74,8 @@ let erase_body_regions (crate : crate) (f : fun_decl) : fun_decl =
               };
           }
         in
-        Some { body with body = erase_visitor#visit_block 0 body.body }
-    | None -> None
+        Body { body with body = erase_visitor#visit_block 0 body.body }
+    | other -> other
   in
 
   let f : fun_decl = { f with body } in
@@ -135,10 +135,10 @@ let remove_unreachable (crate : crate) (f : fun_decl) : fun_decl =
     end
   in
   match f.body with
-  | None -> f
-  | Some body ->
+  | Body body ->
       let body = { body with body = visitor#visit_block () body.body } in
-      { f with body = Some body }
+      { f with body = Body body }
+  | _ -> f
 
 (** The Rust compiler generates a unique implementation of [Default] for arrays
     for every choice of length. For instance, if we write:
@@ -642,8 +642,8 @@ let update_loops (crate : crate) (f : fun_decl) : fun_decl =
   (* Map  *)
   let body =
     match f.body with
-    | Some body -> Some { body with body = visitor#visit_block 0 body.body }
-    | None -> None
+    | Body body -> Body { body with body = visitor#visit_block 0 body.body }
+    | other -> other
   in
 
   let f : fun_decl = { f with body } in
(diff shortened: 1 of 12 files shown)


N1ark:
@sonmarcho CI fails, could you explain to me why I am very unfamiliar with all this 🫠

sonmarcho:
This PR introduced a bug in the extraction: Aeneas now generates `FunsExternal_Template` files in situations when there are no opaque/axiomatized function definitions. As the corresponding `FunsExternal` files (with hand-written models for these opaque functions) are missing, build fails.

src/llbc/LlbcAstUtils.ml
```diff
@@ -37,7 +49,7 @@ let crate_get_opaque_non_builtin_decls (k : crate) (filter_builtin : bool)
   let open ExtractBuiltin in
   let ctx = Charon.NameMatcher.ctx_from_crate k in
   let is_opaque_fun (d : fun_decl) : bool =
-    d.body = None
+    (not (body_is_known Opaque))
```
Nadrieril: Shouldn't this be `(not (body_is_known d.body))`?

N1ark: oops .. yes

src/llbc/LlbcAstUtils.ml
```diff
@@ -15,6 +15,18 @@ end
 module FunIdMap = Collections.MakeMap (FunIdOrderedType)
 module FunIdSet = Collections.MakeSet (FunIdOrderedType)
 
+let body_as_body (f : fun_body) : block gexpr_body option =
+  match f with
+  | Body body -> Some body
+  | _ -> None
+
+let body_as_body_exn f = Option.get (body_as_body f)
```
sonmarcho: What does "exn" mean?

Also I avoid calling `Option.get` directly because it raises exceptions that Aeneas doesn't catch, meaning the translation aborts. For this function, it is better to add `file` and `line` parameters and wrap the calls inside the `[%add_loc]` macro. I note that you used this function to replace calls to `Option.get` that I forgot to properly wrap, but if you could make this modification before forgetting about we forget about these call sites that would be helpful. :)

src/PrePasses.ml
```diff
@@ -1889,15 +1889,16 @@ let apply_passes (crate : crate) : crate =
...
+    with CFailure e ->
       (* The error was already registered, we don't need to register it twice.
          However, we replace the body of the function, and save an error to
          report to the user the fact that we will ignore the function body *)
       let fmt = Print.Crate.crate_to_fmt_env crate in
       let name = Print.name_to_string fmt f.item_meta.name in
       [%save_error] f.item_meta.span
         ("Ignoring the body of '" ^ name ^ "' because of previous error");
-      { f with body = None }
+      let msg = Fmt.to_to_string pp_cfailure e in
```
sonmarcho: You should use `Errors.format_error_message_with_file_line`

src/Translate.ml
```diff
@@ -224,7 +223,8 @@ let translate_function_to_pure_aux (trans_ctx : trans_ctx)
             ctx
         in
         { ctx with forward_inputs }
-    | _ -> [%craise] fdef.item_meta.span "Unreachable"
+    | _, None -> ctx
```
sonmarcho: We should error in the case `Body _, None`

src/llbc/LlbcAstUtils.ml
N1ark: sure! and `_exn` is a suffix in OCaml to indicate a function that raises an exception when there's a non-raising counterpart, i think it's mostly used in JaneStreet libs but i find it quite nice :)

N1ark:
@sonmarcho addressed the comments! any other remarks or is this mergeable?:3

sonmarcho (approved):

sonmarcho:
Sorry about that: I just approved :)

PR #921, opened 2026-04-09
Release pre-compiled Lean library in addition to binaries

joshlf:
This will [allow Hermes to ship these to users](https://github.com/google/zerocopy/issues/3233), allowing users to avoid rebuilding Aeneas when performing verification.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,6 +80,48 @@ jobs:
           nix build -L .#${{ matrix.nix_attr }}
           cp result ${{ matrix.artifact_name }}.tar.gz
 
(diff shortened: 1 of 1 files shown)


joshlf:
Context: https://github.com/google/zerocopy/issues/3233

protz:
This seems to make sense, but I don't know enough about the nix build to approve this.

A question: why can this not be done with the regular `leanprover/lean-action@v1` GitHub action? Is there any way to reuse that?

Nadrieril:
Had a quick look at the lean action, doesn't feel like it would add much here. The interaction with nix is fine.

PR #925, opened 2026-04-10
feat(lean): support overflowing arithmetic operations (add, sub, mul, div)

lotzk:
## Summary

Adds Lean library support for Rust's overflowing arithmetic operations:

- `overflowing_add` (wrapping addition with overflow flag, already existed)
- `overflowing_sub` (wrapping subtraction with overflow flag, new)
- `overflowing_mul` (wrapping multiplication with overflow flag, new)
- `overflowing_div` (division returning with overflow flag for signed integers, new)

All implementations are consistent with Rust semantics and the existing wrapping counterparts (although not formally proved).

## Changes

**`backends/lean/Aeneas/Std/Scalar/Core.lean`**
- Added canonical `UScalar.zero`, `UScalar.one`, `IScalar.zero`, `IScalar.one` constants
  together with their `_bv` characterisation lemmas, facilitating theorems for the new operation.

**`backends/lean/Aeneas/Std/Scalar/OverflowingOps/`** (new files)

`OverflowingOps.lean` is refactored from a single file (that previously only included  `overflowing_add`) into four per-operation files, consistent with the other arithmetic operations. Each file contains:

- Generic `UScalar`/`IScalar` definitions
- `uscalar`/`iscalar`/`core.num.«%S»` aliases
- A `@[step_pure]` spec theorem (`overflowing_*_eq`) for `UScalar` types
- Additional simp lemmas: commutativity, associativity and some other identities (where applicable)

**Note:** `IScalar` step theorems are not included yet but I may add them at a later point.

**`src/extract/ExtractBuiltin.ml`**
- Registers `overflowing_sub`, `overflowing_mul`, and `overflowing_div` for extraction
  (`overflowing_add` was already registered).

## Test cases

A new Rust test file `tests/src/overflowing-ops.rs` covers all four operations for every
integer type (u8 - u128/usize and i8 - i128/isize). The generated Lean extraction is checked in as `tests/lean/OverflowingOps.lean`.


sonmarcho:
Thanks for this! Just to inform you: I'm currently flooded with work, so it will take me a bit of time to review all the PRs that have been stacked... But this PR is definitely a welcome addition.

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
```diff
@@ -0,0 +1,136 @@
...
+namespace Aeneas.Std
+
+open ScalarElab
+
+/-!
+# Overflowing Addition
+-/
+
+def UScalar.overflowing_add {ty} (x y : UScalar ty) : UScalar ty × Bool :=
+  (⟨ BitVec.ofNat _ (x.val + y.val) ⟩, 2^ty.numBits ≤ x.val + y.val)
```
sonmarcho: Now that I review this I think I made a very bad formalisation choice initially (and Mathlib probably evolved quite a bit): I think it would be simpler to use the `BitVec` addition and [`BitVec.uaddOverflow`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Basic.html#BitVec.uaddOverflow).

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
```diff
@@ -0,0 +1,136 @@
...
+
+def IScalar.overflowing_add {ty} (x y : IScalar ty) : IScalar ty × Bool :=
+  (⟨ BitVec.ofInt _ (x.val + y.val) ⟩,
+     ¬ (-2^(ty.numBits -1) ≤ x.val + y.val ∧ x.val + y.val < 2^(ty.numBits-1)))
```
sonmarcho: Similar remark as above

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
```diff
@@ -0,0 +1,126 @@
...
+
+def UScalar.overflowing_div {ty} (x y : UScalar ty) : Result (UScalar ty × Bool) :=
+  if y.bv != 0 then ok (⟨ BitVec.udiv x.bv y.bv ⟩, false) else fail divisionByZero
+
+def IScalar.overflowing_div {ty} (x y : IScalar ty) : Result (IScalar ty × Bool) :=
+  if y.val != 0 then
+    -- There can be an overflow if `x` is equal to the lower bound and `y` to `-1`
+    if ¬(x.val = IScalar.min ty && y.val = -1) then ok (⟨ BitVec.sdiv x.bv y.bv ⟩, false)
+    else ok (x, true)
+  else fail divisionByZero
```
sonmarcho, line 18: I note you're using `BitVec.sdiv` here: we might as well want to use `[BitVec.sdivOverflow](https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Basic.html#BitVec.sdivOverflow)`?

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
```diff
@@ -0,0 +1,136 @@
...
+
+@[simp]
+theorem IScalar.overflowing_add_zero {ty} (x : IScalar ty) :
+  (overflowing_add x IScalar.zero) = (x, false) := by
+  simp [overflowing_add, hmax, hmin]
+
+uscalar @[simp]
+theorem core.num.«%S».overflowing_add_zero(x : «%S») :
+  (overflowing_add x UScalar.zero) = (x, false) :=
+  UScalar.overflowing_add_zero x
```
sonmarcho: Does this lemma trigger on something like `overflowing_add x 0#u32` (which is typically the kind of expressions that would be generated by Aeneas)? It seems you did things right, but it's probably worth adding a test. Also, what about adding the `zero_add` equivalent?

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
```diff
@@ -0,0 +1,126 @@
...
+  overflowing_div x zero = .fail divisionByZero := by
+  simp[overflowing_div]
+
+uscalar @[simp]
+theorem core.num.«%S».overflowing_div_zero(x : «%S») :
+  overflowing_div x UScalar.zero = .fail divisionByZero := UScalar.overflowing_div_zero x
+
+iscalar @[simp]
+theorem core.num.«%S».overflowing_div_zero(x : «%S») :
+  overflowing_div x IScalar.zero = .fail divisionByZero := IScalar.overflowing_div_zero x
```
sonmarcho: Are these actually useful in the context that we're always proving at least panic freedom?

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Sub.lean
```diff
@@ -0,0 +1,114 @@
...
+
+uscalar @[simp]
+theorem core.num.«%S».overflowing_sub_zero(x : «%S») :
+  (overflowing_sub x UScalar.zero) = (x, false) :=
+  UScalar.overflowing_sub_zero x
+
+iscalar @[simp]
+theorem core.num.«%S».overflowing_sub_zero(x : «%S») :
+  (overflowing_sub x IScalar.zero) = (x, false) :=
+   IScalar.overflowing_sub_zero x
```
sonmarcho: Adding `zero_sub` would be useful :)

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Mul.lean
```diff
@@ -0,0 +1,145 @@
...
+  (overflowing_mul x IScalar.one) = (x, false) := by
+  simp [overflowing_mul, IScalar.one, hmax, hmin, one_bv]
+
+uscalar @[simp]
+theorem core.num.«%S».overflowing_mul_one(x : «%S») :
+  (overflowing_mul x UScalar.one) = (x, false) := UScalar.overflowing_mul_one x
+
+iscalar @[simp]
+theorem core.num.«%S».overflowing_mul_one(x : «%S») :
+  (overflowing_mul x IScalar.one) = (x, false) := IScalar.overflowing_mul_one x
```
sonmarcho, line 167: Adding `one_mul` would probably be useful :)

sonmarcho (commented):
This looks good and it seems it took a bit of work: thanks again!
I essentially have two groups of remarks:

- I'm not happy with the way I initially formalized the overflowing operations: I think we should change that (does that seem reasonable?)
- it would be useful to add the symmetric of some of the lemmas you already added

lotzk:
@sonmarcho thanks for the feedback, that all sounds sensible to me. I was not aware of the existing BitVec overflow checks, but I agree these are the right things to use here. I’ll address your comments once I have some more time!

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
lotzk: You are right to question that, these were only helpful in a different context I'm working on. I'm not sure if they are useful in general (and also, they are kind of trivial). Should I remove them, wdyt?

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Sub.lean
lotzk: Added `zero_sub` (AI disclosure: the proof for signed `zero_sub` was generated by Claude).

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Mul.lean
lotzk: Added `one_mul` alongside with `zero_mul`

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
lotzk: Good call, I did not check that at first. I added some `examples` add the end of each file to check the simp lemmas trigger as expected!

lotzk: And also added `zero_add` :)

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
lotzk: Changed the definition to use `BitVec. BitVec.sdivOverflow`. The proofs are a bit more involved now because the `BitVec.sdivOverflow` relies on Int arithmetic.

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
lotzk: Changed to BV addition using `BitVec.uaddOverflow` for the flag. All other ops also follow the same pattern now (apart from unsigned division, which does not overflow).

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Add.lean
lotzk: See reply above

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
sonmarcho: Where do you need them on your case? If that seems useful generally speaking I will be happy to keep them in the Aeneas stdlib. Otherwise yes, it will probably be better to remove them.

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
```diff
@@ -0,0 +1,129 @@
...
+/-!
+## Spec Theorems
+-/
+
+theorem UScalar.overflowing_div_eq {ty} (x y : UScalar ty) :
+  overflowing_div x y = (·, false) <$> UScalar.div x y
+  := by
+    simp[overflowing_div, UScalar.div]
+    rw [map_eq_bind_pure_comp]
+    split<;>simp[pure]
```
sonmarcho: Small nit: the indentation is not consistent with the indentation used in the other proofs.

tests/lean/OverflowingOps.lean
```diff
@@ -0,0 +1,305 @@
+-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
+-- [overflowing_ops]
+import Aeneas
```
sonmarcho, line 3: This file is new, so you need to update the `lakefile.lean` with a new target for `OverflowingOps`

sonmarcho (commented):
Sorry again for the slow responsiveness...
I just have a few minor comments and this is good to go.

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
lotzk: Yes, found a few more of those. All should be consistent now!

tests/lean/OverflowingOps.lean
lotzk: Added

sonmarcho:
Thank for these updates: I resolved all conversations but the one about `overflowing_div_zero`.

backends/lean/Aeneas/Std/Scalar/OverflowingOps/Div.lean
lotzk: Deleted them, I don't think they'll be useful in general.

lotzk:
Thanks! I missed your comment on the `overflowing_div_zero` lemmas. I've deleted these now.

sonmarcho (approved):
I'm merging it once the CI is back, thanks!

PR #933, opened 2026-04-14
[Lean]: Support BITS constant on integer types

lotzk:
## Summary

All Rust integer types provide the associated constant [`BITS`](https://doc.rust-lang.org/std/primitive.u8.html#associatedconstant.BITS), returning the size of the type in bits. Translation does not reduce these constants and instead generates opaque axioms, e.g., using `u8::BITS` generates

```lean
@[rust_const "core::num::{u8}::BITS"] axiom core.num.U8.BITS : Result Std.U32
```

This PR registers `BITS` as a builtin global for all integer types mapping it to 

```lean
UScalar.ofNat (UScalarTy.numBits <Ty>)
```

## Changes

- Register BITS as a builtin global in the extractor (`ExtractBuiltin.ml`) (analogously to  `MIN` and `MAX`)
- Add `core.num.{U8,... , U128, Usize, I8, ..., I128, Isize}.BITS` as reducible definitions in `Scalar/Core.lean`
- Add test functions `u32_use_bits` / `i32_use_bits` in `tests/src/scalars.rs`


sonmarcho:
Thanks for this!

Issue #903, opened 2026-04-02
`Ord` instance for local type placed in Funs.lean but referenced from Types.lean


oliver-butterley:
When a locally-defined type with derived `Ord` is used as the key type of `sorted_vec::SortedSet`, the `SortedSet` field in a struct definition lands in `Types.lean` and references the `Ord` instance. But the `Ord` instance for the local type lands in `Funs.lean`. Since `Funs.lean` imports `Types.lean`, this doesn't compile.

For built-in types like `u16`, the `Ord` instance comes from the Aeneas standard library and is already in scope, so the issue doesn't arise.

## MWE

```toml
# Cargo.toml
[package]
name = "funs-types-dep-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
sorted-vec = "0.8"
```

```rust
// src/lib.rs
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
pub struct Point {
    pub x: u16,
}

pub struct Decoder {
    pub points: sorted_vec::SortedSet<Point>,
}

pub fn new_decoder() -> Decoder {
    Decoder {
        points: sorted_vec::SortedSet::new(),
    }
}
```

```bash
charon cargo --preset=aeneas --hide-marker-traits
aeneas -backend lean -split-files -dest . -subdir Output funs_types_dep_repro.llbc
```

Extraction succeeds with zero errors but the Lean project has the dependency problem.

## The generated output

**`Types.lean`** contains the `Decoder` type definition:

```lean
structure Decoder where
  points : sorted_vec.SortedSet Point.Insts.CoreCmpOrd
```

The field type `sorted_vec.SortedSet` is parameterized by `Point.Insts.CoreCmpOrd` — the derived `Ord` trait instance for `Point`.

**`Funs.lean`** contains the trait instance:

```lean
def Point.Insts.CoreCmpOrd : core.cmp.Ord Point := {
  ...
}
```

## Motivation

Issue encountered in [SparsePostQuantumRatchet](https://github.com/signalapp/SparsePostQuantumRatchet). 
The `PolyDecoder` struct has a field of type `sorted_vec::SortedSet<Pt>`, where `Pt` is a local type with derived `Ord`.

The current workaround is a post-extraction tweak that moves the `Pt` trait instances (`PartialEq`, `Eq`, `PartialOrd`, `Ord`) from `Funs.lean` into `Types.lean` using regex-based find/replace.

## Fix?

[...]

PR #935, opened 2026-04-15
Fix scalar PartialOrd builtin name: PartialCmp -> PartialOrd

mennanov:
The Lean extraction emitted `core.cmp.impls.PartialCmpU64` (and other scalar types) instead of `core.cmp.impls.PartialOrdU64`, causing a name mismatch with the Lean library which defines these via the `scalar` macro as `core.cmp.impls.PartialOrd'S.*` in `Aeneas/Std/Scalar/EqOrd.lean`.

This broke any crate that derives `PartialOrd` on a newtype wrapping a scalar :

```rust
#[derive(PartialOrd)] 
struct View(u64)
```

 Summary of changes:

- Fixes a typo in the builtin name mapping for scalar `PartialOrd` methods in `ExtractBuiltin.ml`: the Lean name prefix was `PartialCmp` instead of `PartialOrd`
- The compiler emitted names like `core.cmp.impls.PartialCmpU64.partial_cmp`, but the Lean library defines them as `core.cmp.impls.PartialOrdU64.partial_cmp`, causing unresolved name errors
- Adds a regression test (`u64_partial_cmp`) that exercises the scalar `PartialOrd` builtin path

Partially addresses #903 (fixes the scalar `PartialOrd` naming, but the `Types.lean`/`Funs.lean` dependency ordering is a separate issue).

### DIFF
--- a/src/extract/ExtractBuiltin.ml
+++ b/src/extract/ExtractBuiltin.ml
@@ -812,7 +812,7 @@ let mk_builtin_funs () : (pattern * Pure.builtin_fun_info) list =
             "core::cmp::impls::{core::cmp::PartialOrd<" ^ ty ^ "," ^ ty ^ ">}::"
             ^ fun_name)
           (fun ty fun_name ->
-            "core.cmp.impls.PartialCmp"
+            "core.cmp.impls.PartialOrd"
             ^ StringUtils.capitalize_first_letter ty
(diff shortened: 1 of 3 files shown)


sonmarcho (approved):
Looks good, thanks!

sonmarcho (approved):

PR #940, opened 2026-04-15
fix: Add default field values for PartialOrd and Ord structures

mennanov:
When Rust derives `PartialOrd`/`Ord`, only `partial_cmp`/`cmp` are generated, the rest use trait defaults. Charon's LLBC only includes the generated methods, so Aeneas emits struct literals missing the default fields. Previously this caused "Fields missing" errors in Lean.

The fix follows the same pattern already used by `PartialEq.ne` and `Eq.assert_receiver_is_total_eq`: Lean fills in defaults when the compiler omits them.

Stacked on #935 (fails to build without it).

### DIFF
--- a/backends/lean/Aeneas/Std/Core/Cmp.lean
+++ b/backends/lean/Aeneas/Std/Core/Cmp.lean
@@ -32,86 +32,113 @@ attribute
   (body := .enum [⟨"Less", "lt", none⟩, ⟨"Equal", "eq", none⟩, ⟨"Greater", "gt", none⟩])]
   Ordering
 
-@[rust_trait "core::cmp::PartialOrd" (parentClauses := ["partialEqInst"])]
-structure core.cmp.PartialOrd (Self : Type) (Rhs : Type) where
-  partialEqInst : core.cmp.PartialEq Self Rhs
-  partial_cmp : Self → Rhs → Result (Option Ordering)
-  lt : Self → Rhs → Result Bool := fun x y => do
-    let cmp ← partial_cmp x y; ok (cmp = some .lt)
-  le : Self → Rhs → Result Bool := fun x y => do
-    let cmp ← partial_cmp x y; ok (cmp = some .lt ∨ cmp = some .eq)
-  gt : Self → Rhs → Result Bool := fun x y => do
-    let cmp ← partial_cmp x y; ok (cmp = some .gt)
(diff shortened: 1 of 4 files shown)


Nadrieril:
Charon's LLBC actually includes default methods if they're used anywhere in the crate, or if the trait in question is explicitly marked as to-include. So you should be able to pass `--include core::cmp::PartialOrd` in order for Charon to always give you all the methods for that trait. (`--translate-all-methods` does it for all traits but that's undesireable because Iterator). If that doesn't work please open a Charon issue.

mennanov:
Thanks for the suggestion! 
I am making my first steps in trying to understand Charon + Aeneas, so my reasoning might be off.

I've tried passing `--include core::cmp::PartialOrd` (and `--include core::cmp::Ord`) to Charon (build-2026.04.15.180725-fbd54169) with Aeneas build-2026.04.15.082433-8ab2e7e4 in my project that uses Charon + Aeneas, but it doesn't work: the default method bodies (`lt`, `le`, `gt`, `ge`) in `core::cmp` use function types internally. Aeneas emitted "Arrow types are not supported yet" errors. The struct literals do get populated in the partial output, but the build fails.

Given that a similar pattern (taken in this PR) is used in `PartialEq.ne` and `Eq.assert_receiver_is_total_eq` do you think this should still be fixed at the Charon level?

Thanks!

Nadrieril:
Ah, so the issue is that the real function bodies present in the standard library are doing something that Aeneas doesn't support (the bodies look like `self.partial_cmp(other).is_some_and(Ordering::is_gt)` which uses a function as a closure). It would be more principled to support this (I opened https://github.com/AeneasVerif/charon/issues/1101 in Charon that should help), but in the mean time your manual fix could very well make sense; I'll let @sonmarcho decide here.

sonmarcho:
Thanks for this! I actually bumped into the same issue recently but did not find the time to wrap up a PR.
I just have one request: would you mind introducing auxiliary functions for `lt`, `le`, etc. rather than writing their bodies directly in `core.cmp.PartialOrd` and `core.cmp.Ord `?

Also, don't hesitate to open an issue with a MWE for the "Arrow types are not supported yet" error: that would be extremely helpful.

sonmarcho (approved):
Thanks for these last changes!

PR #942, opened 2026-04-15
Fix inductive constructor return types with implicit params

mennanov:
## Problem

When a type parameter only appears in a trait bound (e.g. `A: Clone`), Aeneas marks it as implicit `{A : Type}` in Lean. The constructor return types were incorrectly listing this implicit `A` instead of the explicit trait clause parameter.

### Minimal failing example

```rust
use std::marker::PhantomData;

pub enum Holder<K, A: Clone> {
    Value(K),
    Phantom(PhantomData<A>),
}
```

**Before:** constructors returned `Holder K A` (implicit param leaked into return type)
**After:** constructors return `Holder K corecloneCloneInst` (explicit trait clause param)

## Fix

In `extract_type_decl_variant`, filter constructor return type params using `explicit_info` to only include explicit type/const generic params, and append trait clause params.


sonmarcho:
Charon was supposed to filter trait bounds from type declarations: we fixed that in https://github.com/AeneasVerif/charon/pull/1105
Does the issue still persist?

mennanov:
You're right! Updating Charon's version resolved the issue in my project https://github.com/mennanov/simplex/commit/747c446d299d5d650f54aa1ecc3d999b56e1cad6#diff-e14a1184af3607fbd9c4c121352ef738664197c51eed3cd929935e09f1d45098L205-L210 

It looks like this PR is irrelevant.
Thanks!

PR #939, opened 2026-04-15
fix(mvcgen): Allow divergence for partial correctness

abentkamp:
Hello!

Your `WP` instance for `mvcgen` has the same flaw that we had until recently in hax. By assigning `| .div => PredTrans.const ⌜False⌝`, you are essentially saying that divergence can never be considered correct, even when using partial correctness. 

Here is an example attempt of using partial correctness on `div`:

```
example : ⦃ ⌜ True ⌝ ⦄ (Result.div : Result Unit) ⦃ ⇓? r => ⌜ True ⌝ ⦄ := by
  mvcgen
  simp [wp]
  sorry -- Goal is False
```
[...]

PR #944, opened 2026-04-16
feat(mvcgen): specs for +,-,*

abentkamp:
Based on my other PR #939, this PR adds mvcgen specs for addition, subtraction, and multiplication.

I'd like to add more `mvcgen` specs, but maybe we should discuss first whether this is the right way to go.

To state the specifications, I use an abstract postcondition `Q`, which allows me to define the specification in the most general way: for total correctness, partial correctness and everything in between. The Lean developer also use this style [here](https://github.com/leanprover/lean4/blob/714601baf118066cbf3f282361339c6d06665b2a/src/Std/Do/Triple/SpecLemmas.lean) (except that I use the unicode notation for triples).

For the preconditions, it's more convenient to state them as assumptions of the lemma (especially when there is more than one), and just use `⌜ True ⌝` as the Triple precondition. To `mvcgen`, this is equivalent. The Triple precondition is only relevant for stateful monads.

A further discussion point is whether we should mark the theorems as `spec` or maybe only `scoped spec` or maybe some custom attribute, so that users have the chance to provide their own `mvcgen` specs if they wish. A related point is that I eventually would like to have one set of specs for integer reasoning and one set of specs for bitvector reasoning. But maybe just using `spec` for the integer specs is okay to get started?

### DIFF
--- a/backends/lean/Aeneas/Std/Scalar/Core.lean
+++ b/backends/lean/Aeneas/Std/Scalar/Core.lean
@@ -582,7 +582,7 @@ theorem UScalar.tryMkOpt_eq (ty : UScalarTy) (x : Nat) :
 theorem UScalar.tryMk_eq (ty : UScalarTy) (x : Nat) :
   match tryMk ty x with
   | ok y => y.val = x ∧ inBounds ty x
-  | fail _ => ¬ (inBounds ty x)
+  | fail .integerOverflow => ¬ (inBounds ty x)
   | _ => False := by
   have := UScalar.tryMkOpt_eq ty x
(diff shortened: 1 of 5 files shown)


sonmarcho:
We need to discuss this: I'm not in favor of adding theorems specifically for mvcgen as we will end up duplicating a lot of work. We should instead update the `step` attribute so that it also registers an `mvcgen` theorem: this is hopefully a limited change that will give you mvcgen theorems for the whole library almost for free. Shall we discuss this on the Zulip?

abentkamp:
Closing in favor of #950

PR #946, opened 2026-04-17
fix: Add the missing `ne` model for shared-reference PartialEq impl

mennanov:
This PR fixes the following build error:

```
error: Unknown identifier `Shared1A.Insts.CoreCmpPartialEqShared0B.ne`
error: failed to synthesize instance of type class
  Decidable b1
```

When Rust user code does `a != b` on two `&T` values where `T: PartialEq`, the `!=` dispatches through the blanket `impl<A: ?Sized, B: ?Sized> PartialEq<&B> for &A` impl. 

Aeneas was looking up `core::cmp::impls::{core::cmp::PartialEq<&'a @A, &'b @B>}::ne`, finding no builtin mapping, and emitting an opaque axiom with an auto-generated name like `Shared1A.Insts.CoreCmpPartialEqShared0B.ne` that did not exist in the Lean library, producing an "Unknown identifier" build error.

Only the corresponding `.eq` was registered in `backends/lean/Aeneas/Std/Core/Cmp.lean` — the `.ne` method was missing.

### DIFF
--- a/backends/lean/Aeneas/Std/Core/Cmp.lean
+++ b/backends/lean/Aeneas/Std/Core/Cmp.lean
@@ -146,10 +146,16 @@ def core.cmp.impls.PartialEqShared.eq {A : Type} {B : Type} (PartialEqInst : cor
   (x : A) (y : B) : Result Bool :=
   PartialEqInst.eq x y
 
+@[simp, rust_fun "core::cmp::impls::{core::cmp::PartialEq<&'a @A, &'b @B>}::ne"]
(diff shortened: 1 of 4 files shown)


sonmarcho (approved):
Thanks for this! I will merge it once CI is green.

sonmarcho:
Regarding the nix error: I think you need to regenerate the tests (there are diffs in the spans that appear in the comments of the generated Lean code - this often happens when reformatting the Rust code but forgetting to regenerate the Lean).

PR #947, opened 2026-04-19
WIP: add many models extending the Aeneas standard library coverage

oliver-butterley:
The goal is to add to the Aeneas standard library models for everything which is required for curve25519-dalek, libsignal and SparsePostQuantumRatchet.

To make this easier to review, this can be split into separate PRs for each of the different groups of models.

Further work in progress to make it very easy for reviews to see easily that the added models are correct. 

Progress is tracked in  funs-external-stdlib.md.

| Group | Items | Key additions |
|---|---:|---|
| Foundation | 3 | `char` (Lean native `Char`), `IntoIter<T, N>`, `Array::into_iter` |
| Option methods | 5 | `as_ref`, `ok_or`, `Clone`, `Default`, `PartialEq` |
| Result methods | 6 | `is_ok`, `is_err`, `unwrap_or`, `map`, `map_err`, `Try::branch` |
| Vec methods | 8 | `truncate`, `as_slice`, `remove`, `append`, `clear`, `is_empty`, `split_off`, `Default` |
| VecDeque | 4 | type + `new`, `len`, `push_back`, `pop_front` |
| Slice methods | 6 | `clone_from_slice`, `copy_within`, `Ord::cmp`, `concat`, `Concat<[V], T, Vec<T>>`, `PartialEq<[U;N]>` |
| Range | 7 | `Bound<T>`, `RangeFull`, `RangeBounds` trait, bounds impls, `SliceIndex<RangeFull>`, Range::collect/map |
| Iterator adapters | 6 | `Map::next`, `Iterator::map`, collect/map for Range/Iter/IntoIter/Map |
| Array | 1 | `from_fn` with FnMut closure |
| i32 Step trait | 3 | `forward_checked`, `backward_checked`, `steps_between` |
| Cmp/Eq/Borrow | 3 | `Borrow` blanket, `Eq` for u8/usize |
| Misc | 4 | `black_box`, `TryFrom<u64> for u32`, `TryFromIntError`, `str::to_owned` |
| **Deferred** | 2 | `fmt` functions (impure effects) |
| **Total** | **60** | **55 complete** |


oliver-butterley:
There is probably a better way to do this so we can have certainty that the models are all correct in a systematic way...

sonmarcho:
Actually I was open to receiving these models (long term we will generate them from the standard library, but for now we still need them). It would help to split the PR into many small PRs, though. Also, you can use `#[verify::test]`  on the Rust code to generate unit tests: this helps making sure the models are correct.

PR #948, opened 2026-04-20
Prepare for Lean machine integers

abentkamp:
This PR introduces two changes in preparation for migrating the official Lean machine integers:

* I renamed `.val` to `.toNat`/`.toInt` and `.bv` to `.toBitVec`. The goal is to align the naming conventions with ones of the official Lean library. This will make the next step of replacing our custom machine integers with Lean's machine integers easier. In particular, the diffs will be smaller and easier to review.
* I replaced the arithmetic type class instances (+, -, -., *, /, %, >>>, <<<) with custom type class instances (+?, -?, *?, /?, >>>?, <<<?). Once we use the official Lean integers, this will be necessary because otherwise our type class instances will clash with the official Lean instances.

These two changes are separated cleanly into one commit each for easier review of this PR.


abentkamp:
This PR is long out-dated

PR #950, opened 2026-04-21
feat: auto-generate @[spec] mvcgen lemmas from @[step] theorems

abentkamp:
This PR auto-generates an mvcgen @[spec] lemma for every lemma marked with @[step]. This is an alternative approach to https://github.com/AeneasVerif/aeneas/pull/944 that helps us avoid repeating every specification twice (once for step and once for mvcgen).

For now, this PR only allows us to provide mvcgen specs for total correctness. In the future, it would be better to extend the notation to also specify partial correctness specs.


backends/lean/Aeneas/Tactic/Step/Init.lean
```diff
@@ -267,8 +267,40 @@ structure StepSpecAttr where
   ext  : Extension
   deriving Inhabited
 
-private def saveStepSpecFromThm (ext : Extension) (attrKind : AttributeKind) (thName : Name) :
-  AttrM Unit := do
+private def generateMvcgenSpec (stx : Syntax) (attrKind : AttributeKind) (thName : Name) :
+    MetaM Unit := do
+  let env ← getEnv
+  let some decl := env.findAsync? thName
+    | throwError "Could not find theorem {thName}"
```
sonmarcho: `saveStepSpecFromThm` already does this, so I think you can simply add the decl as input to `generateMvcgenSpec`?

backends/lean/Aeneas/Tactic/Step/Init.lean
```diff
@@ -267,8 +267,40 @@ structure StepSpecAttr where
...
+  let sig := decl.sig.get
+  -- Specialise all universe level parameters to 0 so that spec_to_mvcgen {α : Type} applies.
+  -- The generated mvcgen_spec is a valid Type-0 instance of the original theorem.
+  let zeroLevels := List.replicate sig.levelParams.length Level.zero
```
sonmarcho: I think you can generalize and it wouldn't be very hard?
Be aware that I did have universe issues in the past (that is: not all types live in `Type`) so I'm not being overly cautious here.

sonmarcho (changes requested):
Thanks for this! Looks good to me modulo a few minor modifications.

backends/lean/Aeneas/Tactic/Step/Init.lean
abentkamp: Done.

backends/lean/Aeneas/Tactic/Step/Init.lean
abentkamp: I generalized, but what's a bit annoying is that the definition of mvcgen's `WP` forces the `Error` type to live at the same universe level as the return type. So I need to use `(ULift Error)` instead of `Error` now.

sonmarcho (approved):

PR #952, opened 2026-04-21
Fix duplicate phonetic pronunciations and typo

joshtriplett:
The duplicate phonetic pronounciations are a minor thing, but they're the first
thing you see in the README, so they seemed worth fixing.

While fixing that, I noticed a typo, so I fixed that too.


sonmarcho:
Thanks :)

PR #955, opened 2026-04-22
fix: check negativity in UScalar.shiftLeft/shiftRight_IScalar

### LINKED ISSUE #954: Lean implementation does not panic when shifting with negative values
These functions should test whether the value `s` is negative and panic if it is:
```
def UScalar.shiftLeft_IScalar {ty tys} (x : UScalar ty) (s : IScalar tys) :
  Result (UScalar ty) :=
  x.shiftLeft s.toNat

def UScalar.shiftRight_IScalar {ty tys} (x : UScalar ty) (s : IScalar tys) :
  Result (UScalar ty) :=
  x.shiftRight s.toNat
```
https://github.com/AeneasVerif/aeneas/blob/main/backends/lean/Aeneas/Std/Scalar/Bitwise.lean#L37-L43

The corresponding `IScalar.` functions are implemented correctly.

Here is a Rust example to test the behavior in Rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=877349ac304e712a0c8eaf78f75abbb6

abentkamp:
Fixes #954


sonmarcho (approved):
Thanks!

PR #962, opened 2026-04-29
fix: Higher ranked trait bounds

### LINKED ISSUE #765: Bug: Aeneas extraction regression: "Region ids should not be visited directly"
After updating Aeneas from `4301882` to `74749a0` (current `main`) in our [repo](https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify), the extraction step produces 4 errors, all with the same message:

> **Error**: Region ids should not be visited directly; the visitor should catch cases that contain region ids earlier.
> Compiler source: `interp/Interp.ml, line 555`

The affected functions are:

| Function | Source |
|----------|--------|
| `curve25519_dalek::edwards::decompress::step_2` | [`edwards.rs:231-248`](https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify/blob/a071346988dd9474973c1d3b0948c22acccb6514/curve25519-dalek/src/edwards.rs#L231-L248) |
| `curve25519_dalek::montgomery::elligator_encode` | [`montgomery.rs:263-283`](https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify/blob/a071346988dd9474973c1d3b0948c22acccb6514/curve25519-dalek/src/montgomery.rs#L263-L283) |
| `curve25519_dalek::field::FieldElement51::sqrt_ratio_i` | [`field.rs:292-338`](https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify/blob/a071346988dd9474973c1d3b0948c22acccb6514/curve25519-dalek/src/field.rs#L292-L338) |
| `curve25519_dalek::backend::serial::scalar_mul::variable_base::mul` | [`variable_base.rs:11-51`](https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify/blob/a071346988dd9474973c1d3b0948c22acccb6514/curve25519-dalek/src/backend/serial/scalar_mul/variable_base.rs#L11-L51) |

All four functions were extracted successfully with the previous Aeneas version (`43018824`).

### LINKED ISSUE #801: Internal error with higher ranked trait bounds
## Rust reproducer

```rust
trait T<X> {
    fn dummy() -> X;
}

fn test<P: for<'a> T<&'a u8>>() -> u8 {
    *P::dummy()
}
```

## Result

I guess higher ranked trait bounds are not supported but the error message is `[Error] Internal error, please file an issue` so here I am.

### LINKED ISSUE #891: Error with closure taking a reference
## Rust reproducer

```rust
fn f<P: Fn(&usize)> (p: P) {
    p(&1)
}
```

## Aeneas error

`[Error] Internal error, please file an issue`

maximebuyse:
Fixes https://github.com/AeneasVerif/aeneas/issues/801, fixes https://github.com/AeneasVerif/aeneas/issues/891, fixes #765 .
Aeneas crashed with "Internal error" on functions whose generic bounds involve higher-rank trait bounds, e.g. `fn f<P: Fn(&usize)>(p: P)` (the implicit lifetime makes this `for<'a> Fn(&'a usize)`) or `fn test<P: for<'a> T<&'a u8>>()`.

The crash came from a sanity assertion in `generic_args_of_params_erase_regions` that required every trait clause's trait_decl_ref to have empty binder_regions. The fix lifts that restriction: since the function erases all regions in the trait clauses anyway, the bound regions inside the binder are turned into `RErased` and the now-unreferenced `binder_regions` list can simply be cleared.

The codebase has eight further `binder_regions = []` sanity checks that gate other syntactic positions where HRTBs can appear — function pointers, regions_outlive clauses, the trait-method dispatch site, etc. None of those is hit by the reproducers in #891 / #801.


src/llbc/Substitute.ml
```diff
@@ -165,12 +165,23 @@ let generic_args_of_params_erase_regions (span : Meta.span option)
...
      actually, and we only need to erase the regions inside the trait clauses. *)
   let regions = List.map (fun _ -> RErased) generics.regions in
-  (* Erase the regions in the trait clauses. *)
+  (* Erase the regions in the trait clauses.
+
+     Note that the [trait_decl_ref] of a trait clause may have higher-rank
+     regions (HRTBs), e.g. [P : Fn(&usize)] which is sugar for
+     [P : for<'a> Fn(&'a usize)]. Erasing all regions inside the binder makes
+     those bound regions unreferenced, so we can drop the [binder_regions]
+     list afterwards. *)
```
maximebuyse, line 174: I used claude for this fix. The explanation for why dropping `binder_region` instead of checking that it is already empty seems convincing to me. But I have to admit I don't know enough about Aeneas to fully understand the implications.

Nadrieril: I'm afraid any fix will be approximate until such a time where Charon learns to correctly instantiate these higher-ranked bounds

maximebuyse:
Closing as this needs a proper support in charon (https://github.com/AeneasVerif/charon/issues/1143)

PR #963, opened 2026-04-30
fix: `do` elab regressions

mpenciak:
This PR fixes a few regressions from the new `do` elaborator
* Fixes an issue with `step*` not being able to step through binds that use nested patterns
* Fixes an issue with fresh names being generated on pattern matching.
* Add various `uncurry_*` lemmas to the `step_simp` set to simplify proof states when calling `step*`.
* Changes the postcondition elaborator to match the `uncurry` semantics of pattern matching in `do` blocks.
* Fixes the postcondition delaborator to match on the new generated expressions.

Some followup work which is slightly lower priority for the next regression fix:
* The unused variable warnings are not firing on pattern matching binds
* #995


backends/lean/Aeneas/Std/WP.lean
```diff
@@ -204,21 +221,45 @@ scoped syntax:54 term:55 " ⦃ " term " ⦄" : term
...
+/-- Build a `Function.uncurry`-chain wrapping a curried lambda over `xs`. -/
+private partial def buildUncurryLam (xs : List (TSyntax `term)) (body : TSyntax `term) :
+    MacroM (TSyntax `term) := do
+  match xs with
+  | [] => pure body
+  | [x] => `(fun $x => $body)
+  | [a, b] => `(Function.uncurry (fun $a $b => $body))
+  | a :: rest =>
+    let inner ← buildUncurryLam rest body
+    `(Function.uncurry (fun $a => $inner))
```
sonmarcho: I think we can quote `Function.uncurry` in such a way that it is statically resolved (with `mkIdent` maybe)

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -232,36 +273,66 @@ macro_rules
...
-      vars := vars.push { expr := arg, pos := pos.push 0 }
-      pos := pos.push 1
-    k vars { expr := body, pos }
-
 def elabSubExpr (e : SubExpr) : Delab := withTheReader SubExpr (fun _ => e) delab
 
+/-- A post-condition binder slot, as collected by `telescopePredn`. -/
+inductive PostBinder where
+  | single (e : SubExpr)
+  | tuple (es : Array SubExpr)
```
sonmarcho: Do we really need two cases? Or can't we simply check the length of the array (I see you're doing the check in `PostBinder.toTerm`)? Also, does this allow several levels of nesting like in `(((x, y), z), w)` (it seems we only support one level here)?

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -232,36 +273,66 @@ macro_rules
...
+
+/-- Render a `PostBinder` as a `Term`, delabbing each underlying `SubExpr` so
+that variable names line up with how they appear inside the post body. -/
+def PostBinder.toTerm : PostBinder → Delab
+  | .single e => elabSubExpr e
+  | .tuple es => do
+    let ts ← es.mapM elabSubExpr
```
sonmarcho: What happens if the length is 0 (can it happen and do we need to check this case)?

backends/lean/Aeneas/Tactic/Step/Tests/TupleDestruct.lean
```diff
@@ -19,4 +20,39 @@ theorem getPairAndUse_step (x : U32) (h : x.val + 2 ≤ U32.max) :
...
+  ok (a + f + g + d)
+
+/--
+info: Try this:
+
+  [apply]     simp only [step_simps]
+    let* ⟨ a, b, a_post1, a_post2 ⟩ ← twoNats_spec
+    let* ⟨ c, d, c_post1, c_post2 ⟩ ← twoNats_spec
+    let* ⟨ _, _, _ ⟩ ← fourNats_spec
+    let* ⟨ _, _, _ ⟩ ← fourNats_spec
```
sonmarcho: We shouldn't have underscores here

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -582,6 +613,19 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
+     `introN` produces fresh hypotheses we have to give the step_simps lemmas
+     (notably `Function.uncurry_apply_eq`) a chance to fire on them. Without
+     this, nested-tuple posts like `(a, b) (c, d) => …` leave the inner
+     `Function.uncurry (fun c d => …) x_n` un-reduced. -/
+  if posts.size > 0 then
+    withTraceNode `Step (fun _ => pure m!"simpAt: reducing Function.uncurry in posts") do
+    let _ ← Simp.simpAt (simpOnly := true)
+      { maxDischargeDepth := 1, failIfUnchanged := false, iota := false }
+      { simpThms := #[← stepSimpExt.getTheorems] }
+      (.targets posts false)
```
sonmarcho: This is suspicious: there shouldn't be any leftovers by this point. But maybe I need detailed explanations with an example.

backends/lean/Aeneas/Tactic/Step/StepStar.lean
```diff
@@ -395,10 +395,23 @@ def analyzeTarget : TacticM TargetKind := do
...
+        if inner.isLambda then
+          -- Collect ALL binders so multi-binder continuations like
+          -- `Function.uncurry (fun a b => …)` propagate every user name.
+          Lean.Meta.lambdaTelescope inner fun xs _ => do
+            let names ← xs.mapM (·.fvarId!.getUserName)
+            pure (.bind names)
+        else
+          -- Couldn't peek at binder names; the real names will be recovered
+          -- by `tryStep`/`getFirstBind` during the actual step.
+          pure (.bind #[])
```
sonmarcho: This will eventually likely duplicate what `getPostNames` does

tests/lean/TupleTests.lean
```diff
@@ -0,0 +1,51 @@
...
+  let ((e, f), (g, h)) ← bar
+  ok (a + f + g + d)
+
+example : baz ⦃ res => res = 18⦄ := by
+  unfold baz
+  simp only [step_simps]
+  let* ⟨ a, b, a_post1, a_post2 ⟩ ← foo_spec
+  let* ⟨ c, d, c_post1, c_post2 ⟩ ← foo_spec
+  let* ⟨ _, _, _ ⟩ ← bar_spec
+  let* ⟨ _, _, _ ⟩ ← bar_spec
```
sonmarcho: We shouldn't have underscores here

backends/lean/Aeneas/Tactic/Step/StepStar.lean
```diff
@@ -493,8 +506,15 @@ where
...
-      let names := if varName.hasMacroScopes then #[] else #[some varName]
+    | .bind varNames => do
+      -- A binder name shouldn't be used if it is either macro-scoped or synthesized.
+      -- We fall back to post-condition names for unusable names if every name is unusable.
+      -- Otherwise we emit one slot per binder, with `none` for the unusable ones.
+      let unusable (n : Name) : Bool :=
+        n.hasMacroScopes || Step.Name.isElabSynthesized n
+      let names :=
+        if varNames.isEmpty || varNames.all unusable then #[]
+        else varNames.map fun n => if unusable n then none else some n
```
sonmarcho: I'm not sure why we need this. This is probably because some tuples are not properly handled yet.

backends/lean/Aeneas/Std/WP.lean
mpenciak: Yeah this was wrong, `PostBinder` is there to keep track of the shape for when we eventually delaborate. It should be fixed now

backends/lean/Aeneas/Tactic/Step/Step.lean
mpenciak: I think you were right! I was able to remove this by being more careful with the lemmas included in the `introOutputs` call.

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -232,55 +283,130 @@ macro_rules
...
+partial def destructureUncurryChain (slots : Array PostBinder) (body : Expr)
+    (k : Array PostBinder → Expr → Delab) : Delab := do
+  let body' := body.consumeMData
+  match_expr body' with
+  | Function.uncurry _ _ _ f x =>
+    let x := x.consumeMData
+    if x.isFVar then
+      matchUncurryBinders f (fun x1 x2 body => do
+        let s1 := .single { expr := x1, pos := .root }
+        let s2 := .single { expr := x2, pos := .root }
```
sonmarcho: The pos is wrong here

sonmarcho:
I made a few updates:
- I moved some delaborator functions from `Do/Delab.lean` to `Std/Delab.lean` so that they can also be used in the `spec` notation. The spec delaboration is built on top of the do delaboration of patterns.
- I introduced our own version of `uncurry`, to make sure none of the manipulations we do inside of `step` have an impact on user written code
- renamed `predn` to `uncurry'`
- reverted the change in the way post-conditions are named by `step`: the behavior was not consistent with what is expected (when using notation `step ⟨ a, b ⟩` it would introduce names not given by the user)

Remaining issues that we will address in other PRs:
- `HigherOrder.lean`: some of the scripts generated by `step*?` are wrong
- `introOutputs` does not work in a general way, leading to wrong guard messages in `HigherOrder.lean`. We will need to reimplement it - I think that the current approach that consists in instrumenting `simp` to introduce the foralls we want in the order is not powerful enough and we will have to drop it.

I think it's a good idea to be smarter in the way we compute post-condition names when the user doesn't provide any but I prefer to keep that for a subsequent PR.

PR #966, opened 2026-05-04
fix: treat RErased regions as non-outliving

waamm:
Erased regions do not encode lifetime information, so treating them as not outliving the requested region set should match the intended check.

(Noticed this while experimenting with iterators/zip/slices.)

Edit: See [this](https://github.com/AeneasVerif/aeneas/issues/1053) issue.


sonmarcho:
Actually no: when encountering an erased region we have to introduce a fresh region for it and compute the outlive information accordingly. If we get to the line that you modified, it means there is a bug (we did not properly introduce a fresh region before running the analysis). Do you have a MWE for this? If yes, could you open an issue?

PR #970, opened 2026-05-05
fix: rotate_right uses bv.rotateRight, not bv.rotateLeft

remix7531:
UScalar.rotate_right and IScalar.rotate_right were both defined in terms of BitVec.rotateLeft, making them aliases of rotate_left. Any refinement proof for Rust code using rotate_right (SHA-2, ChaCha, BLAKE rotations) would be unprovable.

Also fixes a duplicate `## Rotate Left` heading on the rotate_right section.


sonmarcho (approved):
Thanks! Merging this

PR #971, opened 2026-05-05
feat: Inhabited instance for Array

remix7531:
Default-populates the underlying list via List.replicate when the element type is Inhabited. Useful for proofs that index with `arr[i]!` on arrays of arbitrary element types — `getElem!` requires Inhabited and there was no instance for `Array α n`.


sonmarcho:
Thanks!

PR #975, opened 2026-05-05
Add missing step theorem for unary negation on machine integers

protz:
(gemini provided the last three lines of the proof)

originally found while doing a live-demo with @joshlf 

I am not entirely satisfied with the proof -- consider my self-contained demo example, below.

```lean4
-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
-- [test3]: function definitions
import Aeneas
import Test3.Types
open Aeneas Aeneas.Std Result ControlFlow Error
set_option linter.dupNamespace false
set_option linter.hashCommand false
set_option linter.unusedVariables false

/- You can set the `maxHeartbeats` value with the `-max-heartbeats` CLI option -/
set_option maxHeartbeats 1000000

namespace test3

/-- [test3::abs]:
    Source: 'src/main.rs', lines 1:0-3:1
    Visibility: public -/
def abs (x : Std.I32) : Result Std.I32 := do
  if x < 0#i32
  then -. x
  else ok x

theorem abs_is_abs (x: I32) (h: x ≠ I32.min): abs x ⦃ r => r = |x.val| ⦄ := by
  unfold abs
  step*
  simp [HNeg.hNeg]
  step
  scalar_tac

/-- [test3::main]:
    Source: 'src/main.rs', lines 5:0-5:12 -/
def main : Result Unit := do
  ok ()

end test3
```

1. I assumed it was enough to put the `@[step]` on the IScalar definition, but it looks like I still need to `simp` the negation operator. Why? If I look at subtraction, the `@[step]`s are on the IScalar/UScalar operations, not on the type class for operators.
2. Probably linked to 1., why do I need two calls to step?

Finally, a side-note. While doing the demo with @joshlf, I picked `core.num.I32.min` in the precondition, but it generated a nasty proof obligation that required a ton of `simp` calls. Is it the expectation that those constants should not be used in theorems and `I32.min` and the like behave better for simp theorems? If so, do you want a comment on that? For reference, this was the previous state:

```lean4
theorem abs_is_abs (x: I32) (h: x ≠ core.num.I32.MIN): abs x ⦃ r => r = |x.val| ⦄ := by
  unfold abs
  step*
  simp [HNeg.hNeg]
  step
  . simp [I32.rMin] at h; simp [I32.min,I32.numBits]; assumption
  . scalar_tac
```

Thanks!


protz:
(Note that the proof can also be closed with grind.)

joshlf:
> (Note that the proof can also be closed with grind.)

With or without the theorem added in this PR?

protz:
The proof in this PR can be closed with `grind`. Just pushed a commit to that effect.

If you have this PR applied locally, then the demo example can also be closed with `grind` instead of `scalar_tac`.

sonmarcho:
I think the issue is that `HNeg.hNeg` is not reducible: you need to also add a `step` theorem for it.
With regards to the `simp [I32.rMin]`: this is strange. What is the goal exactly? There might be missing lemmas for signed integers.

protz:
> I think the issue is that `HNeg.hNeg` is not reducible: you need to also add a `step` theorem for it.

no worries, I'll add one too -- this is easy

>  With regards to the simp [I32.rMin]: this is strange. What is the goal exactly? There might be missing lemmas for signed integers.

I had I32.min vs. I32.rMin in hypothesis / goal and the way to close this goal was to reduce both definitions to their concrete value. The example is self-contained so you should be able to take a look? let me know if you need any help / want to discuss this live

sonmarcho:
I see what's going on: i think we need to mark `U8.rMax`, etc. as `@[defeq, simp, scalar_tac_simps, simp_scalar_safe, bvify, grind =, agrind =]`. It might break some proofs. I'm ready to merge whenever you add the step theorem for `HNeg.hNeg`.

protz:
Done, let me know if this is what you were hoping for!

sonmarcho (approved):

PR #976, opened 2026-05-05
FYI: Support raw-pointer casts to non-literal pointee types (SHA-NI extraction)

fournet:
**FYI** — sharing for visibility / discussion. Not requesting merge.

Three small commits surfaced while preparing the SymCrypt SHA-2 SHA-NI
verification target.

## 1. `Add @[step] theorem for Array.to_slice_mut with curried postcondition`

Adds a `@[step]` theorem so that `step` can decompose
`Array.to_slice_mut`. Uses the curried `⦃ s lib => ... ⦄` postcondition
form so callers don't need to destructure a tuple.

## 2. `Curried postconditions for all index_mut/to_slice_mut step specs`

Sweep across `Array/ArraySlice.lean`, `Slice.lean`, `Vec.lean` to convert
the existing `index_mut` / `to_slice_mut` step theorems to the curried
form. One Bst test updated.

## 3. `Support raw-pointer casts to non-literal pointee types`

Generalize `CastRawPtr`'s source/target pointee from `literal_type` to
`ty`, so casts whose pointee is an opaque builtin (e.g. `*const __m128i`,
`*const __m256i`) or any other non-scalar type are translated rather than
raising a fatal error.

On the Lean backend this is emitted as a generic `RawPtr.cast` that
leaves the source and target pointee types implicit (Lean infers them
from the surrounding context, e.g. the argument type of
`_mm_loadu_si128`). On other backends we still restrict to literal
scalar pointees as before.

This fixes the OOM observed when extracting SymCrypt's SHA-NI fast path
(`sha2::sha2_impl::append_blocks_256_shani`), which contains many
`*const u8 as *const __m128i` casts via primitives like:

```rust
unsafe fn loadu_u32xn<const N: usize>(arr: &[u32; N], at: usize)
    -> __m128i {
    _mm_loadu_si128(arr.as_ptr().add(at) as *const __m128i)
}
```

Validated end-to-end: extraction of the full SymCrypt SHA-2 module
including the SHA-NI fast path now produces clean Lean (`def`s for the
load/store primitives, axiomatised intrinsics for `_mm_sha256rnds2_epu32`
etc.).

## Conflict resolution

The cherry-pick of #3 onto current `origin/main` produced a single
trivial conflict against the recent Charon update (`d9d473d0 Update
Charon`): both branches independently renamed
`Charon.Print.cast_kind_to_string` → `Charon.PrintExpressions.cast_kind_to_string`
in the same error site. Resolution: kept this PR's version, which also
updates the surrounding code path.

### DIFF
--- a/backends/lean/Aeneas/Std/Array/ArraySlice.lean
+++ b/backends/lean/Aeneas/Std/Array/ArraySlice.lean
@@ -250,7 +250,7 @@ theorem Array.index_mut_SliceIndexRangeToUsizeSlice {T : Type} {N : Usize}
     (h : r.end ≤ N) :
     core.array.Array.index_mut (core.ops.index.IndexMutSlice
       (core.slice.index.SliceIndexRangeToUsizeSlice T)) a r
-    ⦃ (s, back) =>
+    ⦃ (s : Slice T) (back : Slice T → Array T N) =>
       s.val = a.val.slice 0 r.end ∧
       s.length = r.end.val ∧
       ∀ s', (back s').val = a.val.setSlice! 0 s'.val ⦄ := by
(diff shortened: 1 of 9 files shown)


remix7531:
Hi! Heads-up that our work likely overlaps. I've formally verified RustCrypto's SHA-256, see [my fork](https://github.com/remix7531/rustcrypto_fv/tree/main/hashes/sha2), against a [literate translation](https://remix7531.com/fips-180-4-lean/) of the FIPS-180-4 spec. Happy to collaborate. Contact: remix7531@mailbox.org.

PR #977, opened 2026-05-05
FYI: Discharge sorry in core.slice.Slice.chunks_exact

fournet:
**FYI** — sharing for visibility / discussion. Not requesting merge.

## What

Discharges the `sorry` in `core.slice.Slice.chunks_exact` (in
`backends/lean/Aeneas/Std/SliceIter.lean`).

The proof goes through a small reusable lemma:

```lean
theorem List.toChunks_length_le {α} (n : Nat) (hn : 0 < n) (l : List α) :
    ∀ c ∈ l.toChunks n, c.length ≤ n
```

shown via an auxiliary `List.toChunks.go_length_le` for the internal
go-loop, then bridged to the public `List.toChunks` through a small
case split on the empty-vs-cons input.

The body of `chunks_exact` is then:

```lean
def core.slice.Slice.chunks_exact {T : Type} (s : Slice T) (chunk_size : Std.Usize) :
  Result (core.slice.iter.ChunksExact T) :=
  if hcs : chunk_size.val > 0 then
    ok ⟨ (s.val.toChunks chunk_size.val).attach.map
            (fun ⟨c, hc⟩ => ⟨c, by
              have h1 : c.length ≤ chunk_size.val :=
                List.toChunks_length_le chunk_size.val hcs s.val c hc
              have h2 : chunk_size.val ≤ Std.Usize.max := by scalar_tac
              omega⟩) ⟩
  else fail .panic
```

The previous `s.len % chunk_size.val = 0` guard is dropped — the old
condition over-restricted compared to Rust's `chunks_exact`, which
accepts any input length.

## Caveat (pre-existing, not addressed here)

`List.toChunks` produces a possibly-shorter trailing chunk when
`s.length` is not a multiple of `chunk_size`. Rust's `chunks_exact`
instead places that trailing data into a separate `.remainder()` field
and yields only exactly-sized chunks via `next()`. The present model
puts everything into `chunks`, so for non-multiple inputs the iterator
will yield a too-short final chunk. That's a modelling gap orthogonal
to this fix — discharging the `sorry` doesn't make it worse. A correct
fix would either:

- restore an `s.len % chunk_size.val = 0` precondition (more
  conservative, breaks Rust call sites with non-multiple input), or
- model the remainder explicitly (extend `ChunksExact` with a
  `remainder : Slice T` field; chunk the prefix to a multiple length).

Happy to follow up either way if you have a preference.

## Context

Surfaced while preparing extraction of the SymCrypt SHA-2 SHA-NI fast
path, which calls `chunks_exact(64)` on byte slices whose length is
guaranteed to be a multiple of 64 by debug_assert at the call site.
The discharge keeps that branch axiom-free.

### DIFF
--- a/backends/lean/Aeneas/Std/SliceIter.lean
+++ b/backends/lean/Aeneas/Std/SliceIter.lean
@@ -138,11 +138,58 @@ def core.iter.traits.iterator.IteratorChunksExact (T : Type) :
   take := core.slice.iter.IteratorChunksExact.take
 }
 
+/-- Auxiliary: every chunk in `List.toChunks.go n xs acc1 acc2` has length ≤ n,
+    given the obvious accumulator preconditions. -/
+private theorem List.toChunks.go_length_le {α} (n : Nat) (hn : 0 < n) :
+    ∀ (xs : List α) (acc1 : Array α) (acc2 : Array (List α)),
+      acc1.size ≤ n →
+      (∀ c ∈ acc2, c.length ≤ n) →
+      ∀ c ∈ List.toChunks.go n xs acc1 acc2, c.length ≤ n := by
+  intro xs
+  induction xs with
+  | nil =>
+    intro acc1 acc2 hsz hacc2 c hc
+    change c ∈ (acc2.push acc1.toList).toList at hc
+    simp at hc
+    rcases hc with hc | hc
+    · exact hacc2 c hc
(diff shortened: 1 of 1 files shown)


sonmarcho:
Closing because subsumed by https://github.com/AeneasVerif/aeneas/pull/998

PR #981, opened 2026-05-06
Add release tag workflow for Lean updates.

oliver-butterley:
This PR adds lean-release-tag.yml workflow. If a push on main changes `backends/lean/lean-toolchain` then the action fires and creates a tag and a github release (named e.g. `v4.30.0-rc2`). 

This is the normality for Lean projects.

Note: the tag/release is always on the first commit for a given version of Lean.


oliver-butterley:
This PR doesn't use the [standard action for such Lean project release tags](https://github.com/leanprover-community/lean-release-tag) since that action currently requires the Lean project to be at root. Consequently the bash logic is inlined here. It would be a simple change to generalise the action and perhaps it will happen soon. ([Issue](https://github.com/leanprover-community/lean-release-tag/issues/3) for this.) 

We could wait until that update happens if it is just a few days.

Nadrieril:
I can't say whether this is a thing we want or not, but if we do I still see two issues with this: the first is that we already use github releases to publish artifacts, this would create releases without the artifacts which seems confusing. The second is that aeneas itself will eventually have versions, so tagging releases with the lean version seems wrong to me.

What's the rationale behind the proposed behavior? Do you expect people will target a specific aeneas release based on the lean toolchain version used?

oliver-butterley:
> What's the rationale behind the proposed behavior? Do you expect people will target a specific aeneas release based on the lean toolchain version used?

I'll add here the comment I wrote directly to Son concerning this PR:

> Note: The standard github action adds a tag to the first commit which uses a given version of Lean. On the other hand, when one specifies a particular version downstream in a lakefile, one would expect to use the latest commit which supports a given version of Lean. Moreover, with Aeneas users will still need to use a specific commit because the extraction machinery will need to be fixed.
> 
> In conclusion, the tags will be convenient for showing updated versions exist but won't be used in practice by anyone. It will fit the normality for Lean projects. (Although I hope the Lean versioning system improves soon!)

I don't see a huge advantage to these tags, I offered to help with this since I know about it and others requested it. When the Lean versioning system is more useful, it would be good to have such for the lean project within the repo.

oliver-butterley:
The real important thing from the Lean side is to regularly update to the latest version of Lean and Mathlib since improvements are being integrated constantly.

Nadrieril:
What's the exact need, I don't think I'm following. Is it being able to know what the toolchain version in use is? Is it knowing when a new toolchain is being used? Am I wrong to think that a good changelog practice would fill that need?

oliver-butterley:
From the Lean side the Lean toolchain is important: Typically we have a Lean project which imports several other Lean projects and this must be done in a way that all the lean projects use the same underlying toolchain. There isn't (yet) a sophisticated version and dependency management system for Lean. 

Tagging when the toolchain changes is standard in Lean projects but doesn't completely solve the real problem. To some extent it helps because then your lean project can important all the others it depends on using the tags. However using Aeneas I would always pin a specific commit of Aeneas and then I would adjust the other imports to match.

My initial point of view was there were no downsides to the tagging, it was standard and [people wanted this]([#Software Verification in Lean Hackathon > Lean version used by Aeneas](https://leanprover.zulipchat.com/#narrow/channel/562866-Software-Verification-in-Lean-Hackathon/topic/Lean.20version.20used.20by.20Aeneas/with/593176154)). Consequently worthwhile. But, if the downsides are significant, better to not do it!

Nadrieril:
Ah, so it's to facilitate synchronized migrations! You can only update the toolchain of your project once all your dependencies have updated, and you use these tags to facilitate updating all your deps. Ok that makes sense to me.

Then I'm still interested in finding solutions to the two problems I raised. For the second, we should name these `lean-v4.30.0-rc2` instead of just `v4.30.0-rc2` imo. For the first, I'm not sure whether having some releases without artifacts is a problem for @joshlf's workflow.

oliver-butterley:
> should name these lean-v4.30.0-rc2 instead of just v4.30.0-rc2 imo.

Yes, much better naming.

> I'm not sure whether having some releases without artifacts is a problem for @joshlf's workflow.

The lean version can be simply a tag and not a release. That might avoid the interference with other workflows.

Don't merge this for the moment and I'll ask in the long thread what people actually want and why. 

Also I suspect soon this can be handled by the standard action for lean project releases and so the code can be a bit tidier.

oliver-butterley:
It seems to me that people who asked for this initially had in mind something slightly different. Anyway nobody continued with any strong enthusiasm. For me it's not useful all the time the Aeneas extraction code is in the same repo as the Aeneas Lean library and all our repos follow the new commits of Aeneas very closely. 

I would suggest closing this PR without merging.

PR #983, opened 2026-05-06
Update to Lean v4.30.0-rc2

oliver-butterley:
This PR updates the Lean projects to toolchain v4.30.0-rc2.

Very few changes required. 

- `Array.getElem!_toList` is now provided by Lean's stdlib (since 4.30 with the same name):
https://leanprover-community.github.io/mathlib4_docs/Init/Data/Array/Basic.html#Array.getElem!_toList

The following two related theroems are now provided by Lean's stdlib. 

- `Array.getElem!_eq_toList_getElem!` is provided by `Array.getElem!_toList` (but reverse direction)
- `Array.getElem_eq_toList_getElem` is provided by `Array.getElem_toList` (but reverse direction)

Each is now marked as deprecated and all the other proofs in the project which used to use them now use the theroems available upstream.


backends/lean/Aeneas/Data/Array.lean
```diff
@@ -10,11 +10,13 @@ attribute [scalar_tac_simps, simp_lists_safe] Array.size
...
+-- TODO: this is `Array.getElem!_toList` but reverse direction
 @[local simp]
 theorem getElem!_eq_toList_getElem! {α} [Inhabited α] (s : Array α) (i : ℕ) :
   s[i]! = s.toList[i]! := by
   cases s; simp
 
+-- TODO: this is `Array.getElem_toList` but reverse direction
 @[local simp]
 theorem getElem_eq_toList_getElem {α} [Inhabited α] (s : Array α) (i : ℕ) (hi : i < s.size) :
   s[i] = s.toList[i] := by
```
sonmarcho: Would something like `local simp <-` on the stdlib theorems work (so that we can get rid of these local theorems altogether)?

oliver-butterley: Yes, technically we could do `attribute [local simp ←] Array.getElem!_toList Array.getElem_toList`. Then relatively minor updates to various other proofs in this file which explicitly mention the previous lemmas.

However it feels generally unhealthy to always be working in the opposite direction to the simp normal form direction established upstream. 

Unless there is a particular reason for Aeneas to operate with a different notion of simp normal form with lists, I think it would be good to bring all of this inline with upstream in the near future. Not critical but it will make things smoother long term.

sonmarcho: I think the lean std theorems are not marked as `simp` (otherwise `simp` would loop in this file)? So I think it's ok if we mark them as `local simp`(as long as it's local it's ok: these theorems are here only to make the proofs in this file go through, they're not part of the global simp set). But if you find a way of removing them (and marking the Lean std ones as `local simp`) that would be even better.

oliver-butterley: No, there are both simp lemmas:
- [`Array.getElem!_toList`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Array/Basic.html#Array.getElem!_toList)
- [`Array.getElem_toList`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Array/Basic.html#Array.getElem_toList)

What is preferred? I've done what seemed best to me:

Marked the definitions of `getElem!_eq_toList_getElem!` and `getElem_eq_toList_getElem` as deprecated and adjusted the proofs in `Array.lean` and `Vector.lean` to use the upstream theorems. Mostly this is just the following replacements since the previous theorems were mostly used with `simp only` calls.)

Find: `getElem!_eq_toList_getElem!`; Replace: `← getElem!_toList`
Find: `getElem_eq_toList_getElem`; Replace: `← getElem_toList`

TODO: 
- Sometime in the future look carefully at Lists/Arrays in Aeneas to see if something can be done to better align it with the upstream simp normal form direction.

sonmarcho: That looks good, thanks :)

PR #1009, opened 2026-05-08
Emit translation.json

oliver-butterley:
Feature: A file called `translation.json` is produced alongside the Lean files.

Guiding principle:

The json describes **what Aeneas did**: the Lean declarations it produced and the connection to the original Rust source, this is the data only available after Aeneas translation. It does not duplicate facts that are already available in the LLBC. Consumers of the data are expected to read those from the `.llbc` and join with the `manifest.json` by `def_id`.

Motivation:

This allows consumers of the data to solidly connect the data concerning spec theorems back to the original Rust source code. Designed to be a minimal and reusable source of data. A full system is likely to need to combine the data from llbc, manifest together with data from the Lean project to give meaniful information on verification.


oliver-butterley:
Further notes/discussion:

In general I see three alternatives for sourcing this data:

-  Scraping Rust info from the Lean docstrings (current situation). This is hacky and fragile.
-  Aeneas outputs a json file together with the Lean files with the info that is only available after translation.
-  Custom lean attributes so Aeneas attaches the relevant info to the Lean definitions that it outputs in a way that can be easily and reliably read.

Both 2 and 3 have clear benefits over the current situation. Option 2 has the benefit that the information can be easily processed in any language whereas option 3 is best suited for using Lean. 

Code is arranged in a way that it gives very minimal disturbance to the core translation code. 

No tests are included in this PR. Are they desired for this type of feature?

documentation/emit-json.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+`llbc_file` is the basename of the `.llbc` input. `lean_files` are relative to the directory containing `translation.json`. `subdir` is omitted when the `-subdir` flag was not used.
+
+### Function entry fields
+
+| Field | Always present | Meaning |
+|---|---|---|
+| `def_id` | yes | `FunDeclId` (join key into `.llbc`) |
+| `lean_id` | yes | Fully-qualified Lean name (`Namespace.BaseName`) |
```
sonmarcho: Can we rename to `lean_name`?

oliver-butterley: Absolutely, I'll do that.

oliver-butterley:
@sonmarcho I made the change you proposed and multiple small improvements. 

Implementation notes: 
The majority of the code is in a single new file which defines `EmitJson`. This defines several hooks which are then called from within `translate.ml` at the key points. Every EmitJson hook is a no-op if the flag `-emit-json` isn't set. (They all have the suffix `_if_enabled` to clarify this but the suffix could be dropped if brevity is preferred). 

In particular I've tried hard to make it convincing that it won't break anything for users that don't add the `-emit-json` flag. 

Question 1:
The outputted data is really minimal, it is strictly what is not obtained from anywhere except Aeneas. I could see the argument for adding a few more fields to the data, specifically:
- The fully-qualified Rust name (item_meta.name)
- The Rust source info, file name, line number (item_meta.span)

This would mean that for a large number of use cases one doesn't need to look at the llbc and can immediately get all the key data from the single json.  What do you think about including these extra fields?

Question 2:
Currently translation.json only includes Type, Fun and Global. It ignores TraitDecl and TraitImpl. Would it be best to include these for completeness?

documentation/emit-json.md
```diff
@@ -0,0 +1,52 @@
...
+```json
+{
+  "aeneas_version": "abc1234",
+  "charon_version": "xyz5678",
+  "crate": "my_crate",
+  "output": {
+    "subdir": "SubDir",
+    "llbc_file": "input.llbc",
+    "lean_files": ["SubDir/Types.lean", "SubDir/Funs.lean"]
+  },
```
sonmarcho: The `llbc_file` is an input and may come from a different folder that `subdir ?

documentation/emit-json.md
```diff
@@ -0,0 +1,52 @@
...
+| `def_id` | yes | `FunDeclId` (join key into `.llbc`) |
+| `lean_name` | yes | Fully-qualified Lean name (`Namespace.Name`) |
+| `lean_file` | yes | Path relative to `dest_dir` |
+| `is_opaque` | yes | Extracted as an axiom (no body in the Pure AST) |
+| `can_fail` | yes | Return type wrapped in `Result` (function can panic) |
+| `can_diverge` | yes | May not terminate |
+| `is_rec` | yes | Part of a mutually recursive group |
+| `reducible` | yes | Marked `@[reducible]` by Aeneas |
+| `loop` | loop entries only | `{ "id": N, "pos": [...], "is_body": bool }` |
+| `parent_lean_name` | loop entries only | `lean_name` of the enclosing Rust function |
```
sonmarcho, line 55: `is_local` would be useful too I think

src/EmitJson.ml
```diff
@@ -0,0 +1,248 @@
...
+let init_if_enabled ~(dest_dir : string) : unit =
+  if !Config.emit_json then begin
+    state.function_entries <- [];
+    state.type_entries <- [];
+    state.global_entries <- [];
+    state.lean_files <- [];
+    state.current_lean_file <- "";
+    state.current_lean_namespace <- "";
+    state.dest_dir <- dest_dir
+  end
```
sonmarcho: I'm surprised: why do we need to reinitialize the state (it is already properly initialized by `make_state`)?

src/EmitJson.ml
```diff
@@ -0,0 +1,248 @@
...
+
+(* ------------------------------------------------------------------------ *)
+(* Entry construction                                                       *)
+(* ------------------------------------------------------------------------ *)
+
+let qualify (basename : string) : string =
```
sonmarcho: It would be useful to add a docstring explaining what this function does and why we need it

sonmarcho (commented):
Very nice and clean PR, thanks! I just have a few comments and questions.

documentation/emit-json.md
oliver-butterley: Yes... several different things here:

Firstly, of course the llbc_file should not be under output, that naming makes no sense. Changing this to field to "files". Also removing the subdir field since that is useless on its own.

The motive here is have a solid way to know where llbc and lean files are relative to the json. Since we require the data from the llbc to have complete info, it is good if we can locate it from this json! I'll update the code so that it makes an attempt to store the relative path to the llbc or absolute path if that can't be done.

documentation/emit-json.md
oliver-butterley: Yes, I think it is convenient to have it here. I'll add it. 

It wasn't added previously because of the strong emphasis on not having anything here that is already available in the llbc and this is the case for is_local. 

On the other hand, I'm in favour of having all the key details that the majority of consumers need already here in this json.

src/EmitJson.ml
oliver-butterley: Indeed, that is superfluous! We only need the last line, `state.dest_dir <- dest_dir`. I'll trim it.

src/EmitJson.ml
oliver-butterley: Yes, it's not so obvious. Also deserves a better name in addition to an explanation! I'll call it `full_lean_name` and add a docstring.

oliver-butterley:
I made various alterations following the comments. Ready for review again!

sonmarcho:
That looks great, thanks! I just realized that I forgot to ask you to add TraitDecl and TraitImpl: could you do that?

oliver-butterley:
> I just realized that I forgot to ask you to add TraitDecl and TraitImpl: could you do that?

Sure thing! They are now included. There is a slight subtlety related to linking to builtin trait impls, let me know if the rationale I've taken makes sense to you (there is a [paragraph in the documentation](https://github.com/oliver-butterley/aeneas/blob/c2ab5c60c28e5bf3dbb786b4a8a7fc41cf736901/documentation/emit-json.md) to explain it). It seems the best option from my point of you! 

Similar to the others, this just adds two hooks into the main code and both are no-op when -emitJson is not set.

sonmarcho (approved):
Perfect, thanks!

PR #1016, opened 2026-05-12
Switch to a nightly release schedule

Nadrieril:
The per-commit tags were causing a lot of noise, nightly feels regular enough and lighter. @joshlf does that work for you?


joshlf:
No objection!

PR #1019, opened 2026-05-12
fix: Correctly flag used and unused variables

### LINKED ISSUE #995: Bug: unused variable warning with the custom do elaborator
When using the custom do elaborator on this code we get a warning:
```lean
def test39 (n : Nat) : Result Nat := do
  match n with
  | 0 => .ok 42
  | k + 1 => -- warning: "unused variable `k`"
    let a ← .ok (k + 10)
    let b ← .ok (a + 1)
    .ok b
```

There is no warning when using the regular do elaborator.

mpenciak:
Fixes #995. 

We need to populate the infotree for the elaborated terms when we introduce binders (both in let bindings, and match arms). This also fixes some unnoticed errors where the unused variable linter should have been firing but wasn't.


backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -435,11 +487,18 @@ partial def elabDoMatch
     (discrs : Array (TSyntax ``Term.matchDiscr))
     (alts : Array (TSyntax ``Term.matchAlt)) (rest : List DoElem) : ElabM Expr := do
   let mut armSeqs : Array (TSyntax ``doSeq) := #[]
+  let mut armPatRanges : Array Syntax.Range := #[]
   let mut holedAlts : Array (TSyntax ``Term.matchAlt) := #[]
   for alt in alts do
     match alt with
     | `(matchAltExpr| | $pats,* => $body) =>
+      let elems := pats.getElems
```
sonmarcho, line 495: I would add an `if elems.size = 0 then throwError "..."` to make sure the access `elems[0]!` below does not panic (panics are hard to debug)

sonmarcho (changes requested):
Looks good, thanks! I only have a minor request.

sonmarcho (approved):

PR #1021, opened 2026-05-13
Annotate bare ADT constructors when Lean cannot infer their type

### LINKED ISSUE #1018: Type information can get lost (and inference is not possible)
The following rust example:

```rust
fn test() -> bool {
    let x: Option<u8> = None;
    x.is_none()
}
```

gets translated to the following lean:

```lean
def test : Result Bool := do
  ok (core.option.Option.is_none none)
```

And here lean fails because it doesn't know how to instantiate the type argument for `option` of the `none` literal.

maximebuyse:
Fixes #1018.

This PR is claude-generated. It directly targets the family of cases covered by the reproducer in the issue. However this means in some other cases it adds the type annotation even if lean would be able to infer (the diff in existing tests shows some examples). This allows to keep the logic quite simple, and hopefully the result is not too polluted by the superfluous annotations.


sonmarcho:
Can you try to give the following to Claude?
Given `f x1 ... xn`, compute the set of type parameters of (the uninstantiated) `f` that are implicit and are not constrained by the arguments of `f x1 ... xn` (the application may be partial - for instance, `a` is constrained in `f x` where `f : {a b} -> a -> b -> unit`, but `b` is not). These type parameters are unconstrained even if we know the types of `x1`, ..., `xn` (and the micro-pass that infers annotation makes sure that types of arguments are known, in a conservative manner). If this set is non-empty and the known type of the application contains holes (we could be more precise by requiring it contains holes for some of the types in the set), then introduce a type annotation for the whole application expression (i.e., `f x1 ... xn : ty`).

maximebuyse:
> Can you try to give the following to Claude? Given `f x1 ... xn`, compute the set of type parameters of (the uninstantiated) `f` that are implicit and are not constrained by the arguments of `f x1 ... xn` (the application may be partial - for instance, `a` is constrained in `f x` where `f : {a b} -> a -> b -> unit`, but `b` is not). These type parameters are unconstrained even if we know the types of `x1`, ..., `xn` (and the micro-pass that infers annotation makes sure that types of arguments are known, in a conservative manner). If this set is non-empty and the known type of the application contains holes (we could be more precise by requiring it contains holes for some of the types in the set), then introduce a type annotation for the whole application expression (i.e., `f x1 ... xn : ty`).

Thanks for the suggestion. I pushed an updated version. Now it doesn't produce the unnecessary annotations. In some cases it also produces less annotations (as can be seen from the diff in test output). But this should be for cases where lean can infer so I would consider it an improvement.

tests/lean/AdtBorrows.lean
```diff
@@ -146,7 +146,7 @@ def MutWrapper1.create
     Source: 'tests/src/adt-borrows.rs', lines 108:4-110:5 -/
 def MutWrapper1.unwrap
   {T : Type} (self : MutWrapper1 T) : Result (T × (T → MutWrapper1 T)) := do
-  let back := fun t => ({ x := t } : MutWrapper1 T)
+  let back := fun t => { x := t }
```
sonmarcho, line 149: This one is a bit suspicious: I need to investigate (the inference algorithm I implemented shouldn't assume we know the type of `back` here, so it shouldn't assume we can infer the type of `{x := t}`). I think there is a bug, but unrelated to what you did.

sonmarcho:
Thanks! I'll need a bit of time to investigate whether the diff is correct, though.

PR #1029, opened 2026-05-13
Add support for multi-target extraction

sonmarcho:
This PR adds support for Charon's multi-target extraction.

# Generated code shape

For a function like:
```rust
fn dispatch_add(a: u128, b: u128) -> u128 {
    #[cfg(target_arch = "x86_64")]
    { add_vec::<x86::Sse2>(a, b) }
    #[cfg(target_arch = "aarch64")]
    { add_vec::<arm::Neon>(a, b) }
}
```

Aeneas generates the following:
```lean
def dispatch_add.x86_64 (a : U128) (b : U128) : Result U128 := do
  ...  -- x86 body

def dispatch_add.aarch64 (a : U128) (b : U128) : Result U128 := do
  ...  -- arm body

def dispatch_add (a : U128) (b : U128) : Result U128 := do
  let tgt ← get_target
  if tgt = (toStr "x86_64")
  then dispatch_add.x86_64 a b
  else dispatch_add.aarch64 a b
```

where `get_target` is an opaque axiom (`axiom get_target : Result Str`) with a trivial
step spec (`get_target ⦃ fun _ => True ⦄`). Nothing meaningful can be deduced from
its output — it only serves as a dispatch mechanism.

# Name suffix stripping

Charon appends a `PeTarget` suffix to all per-target item names for disambiguation.
A prepass (`strip_non_dispatch_target_suffixes`) removes unnecessary suffixes:

- **Functions**: the suffix is stripped when the function is not behind a target
  dispatch (i.e., its `item_source` is not `TargetDependentItem`). These are
  functions that exist for only one target — there is no ambiguity.

- **Types**: a multiset of base names (name without suffix) is computed. If a base
  name appears only once, the suffix is removed (no collision).


Nadrieril:
The name suffix stripping would make perfect sense to be done on the charon side

PR #1033, opened 2026-05-15
Bump Charon

N1ark:
Partner PR to https://github.com/AeneasVerif/charon/pull/1033

### DIFF
--- a/src/pure/PureMicroPassesGeneral.ml
+++ b/src/pure/PureMicroPassesGeneral.ml
@@ -2184,6 +2184,8 @@ let eliminate_box_functions_visitor (_ctx : ctx) (def : fun_decl) =
               | BoxNew ->
                   let arg, args = Collections.List.pop args in
                   [%add_loc] mk_apps span arg args
+              | BoxWrite
+              | SliceIntoVec
               | Index _
               | ArrayToSliceShared
               | ArrayToSliceMut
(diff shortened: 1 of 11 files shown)


src/pure/PureMicroPassesGeneral.ml
N1ark, line 2196: not super convinced by this but it seems to work?...

sonmarcho: I don't fully remember the context here, but I think it works: we eliminated the box type already but preserve the `BoxWrite` and `SliceIntoVec` functions (that is: `BoxWrite` consumes a `MaybeUninit<T>` and "unwraps" it into a `T`, while `SliceIntoVec` converts a Slice into a Vec). At the same time we need Lean models for these functions: it seems none of the tests generates code that uses these. Could you come up with a test that does?

N1ark: No tests generates these because currently Aeneas doesn't support `vec!`, and these builtins only occur when `vec!` is used; I fear I lack the expertise and time to implement support for it in Aeneas, as it requires a bit of extra engineering. would you be happy merging this as-is, as it wouldn't break anything, and adding `vec!` support later down the line?

to add support to it you would need to define the signatures of `BoxWrite` and `SliceIntoVec`; these builtin functions  use `Vec`/`MaybeUninit`, so defining their signature requires access to the crate, which isn't possible right now and would need a bit of extra plumbing :/

Nadrieril: Aeneas does support `vec![]` today tho, any simple `vec![x, y, z]` works fine

Nadrieril:
Superceded by https://github.com/AeneasVerif/aeneas/pull/1075

PR #1045, opened 2026-05-20
fix: allow unit binds

mpenciak:
We didn't support the `() <- f x` notation before. 

This notation is used in `StepStar.lean` and the failure to build the example with the new `do` elaborator wasn't caught, so we add the import earlier in the import graph to apply it to the notation in the inline tests in the library.


backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -206,6 +206,10 @@ partial def analyzePat (pat : Term) (ty : Expr) : ElabM PatShape := do
     subPats.toList.zip subTypes |>.toArray.mapM fun (p, t) => analyzePat p t
   match pat with
   | `(_) => return .leaf none
+  | `(()) =>
+    unless (← whnf ty).isAppOf ``PUnit do
```
sonmarcho, line 210: I think we should also allow ``Unit (can it happen in practice)?

sonmarcho (approved):

backends/lean/Aeneas/Do/Elab.lean
mpenciak: I think because `Unit` is an abbrev of `PUnit`, this handles those cases as well. I tried a bunch of different permutations of syntax/type signatures and this check didn't seem to fail on anything well-formed.

PR #1047, opened 2026-05-21
fix: Correctly destructure fvars based off shape of binding in step call

### LINKED ISSUE #1025: bug: Generalize `Step.introOutputs`
`introOutputs` doesn't work if the way the let binding decompose an output is not consistent with the step theorem that is used. For instance:
```lean
theorem f.step ... : f x ⦃ a b c => ... ⦄ -- 3 outputs

let (a, b) ← f x -- 2 outputs
```

Also, `introOutputs` currently greedily decomposes tuples, which it shouldn't. See: https://github.com/AeneasVerif/aeneas/blob/e13f77a7804d84b511b897c16a15d9c3f4e54cbf/backends/lean/Aeneas/Tactic/Step/Tests/HigherOrder.lean#L43

mpenciak:
Closes #1025 

Instead of relying on the `simp` calls in `introOutputs` to destructure the patterned binders, we use the shape of the pattern to do the manual destructuring. 

In cases that the manual shape is not totally destructured compared to the postcondition, there will still be some `uncurry'`s left, so we add an additional simp pass to `introOutput` to handle that as well.


backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -205,7 +205,8 @@ partial def analyzePat (pat : Term) (ty : Expr) : ElabM PatShape := do
   let analyzeSubs (subPats : Array Term) (subTypes : List Expr) : ElabM (Array PatShape) :=
     subPats.toList.zip subTypes |>.toArray.mapM fun (p, t) => analyzePat p t
   match pat with
-  | `(_) => return .leaf none
+  -- just here until the unit return PR gets merged
+  | `(_) | `(()) => return .leaf none
```
sonmarcho: Putting a comment so that we don't forget to update this (I approved https://github.com/AeneasVerif/aeneas/pull/1045 if this is the PR you're referring to)

backends/lean/Aeneas/Std/Primitives.lean
```diff
@@ -193,6 +193,20 @@ be directly manipulated by the user. -/
...
+
+We restrict the final return type to `Prop` so the simp lemmas cannot fire on
+bind continuations. `uncurry_eq_prop` handles the base case and
+`uncurry_eq_prop_arrow` handles the curried case. -/
+
+theorem uncurry_eq_prop {α β} (x : α × β) (p : α → β → Prop) :
+    uncurry p x = p x.fst x.snd := by cases x; rfl
+
+theorem uncurry_eq_prop_arrow {α β σ} (x : α × β) (p : α → β → σ → Prop) :
+    uncurry p x = p x.fst x.snd := by cases x; rfl
```
sonmarcho, line 208: Is that really general enough? For instance, why don't we need to handle the case `a -> b -> c -> d -> Prop` (we don't want a specific lemma for this of course)?

backends/lean/Aeneas/Tactic/Step/Tests/IntroOutputs.lean
```diff
@@ -0,0 +1,410 @@
...
+  unfold quadProg; step*
+
+/--
+info: Try this:
+
+  [apply]     let* ⟨ a, b, a_post1, a_post2, a_post3, a_post4 ⟩ ← quadProg_spec
+    agrind
+-/
+#guard_msgs in
+example : (do let (a, b) ← quadProg
```
sonmarcho, line 204: Would it be possible to add tests with #guard_msgs showing the environment after the first call to `step`? It would be useful to make sure we get the proper destructuring (here we want to check that `a` and `b` in the spec got substituted with `a.fst` and `a.snd` I think).

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -3,6 +3,8 @@ import Aeneas.Tactic.Solver.ScalarTac
 import Aeneas.Tactic.Step.Init
 import Aeneas.Tactic.Step.GrindState
 import Aeneas.Std
+-- just here until the unit bind PR gets merged
+import Aeneas.Do
```
sonmarcho, line 7: Let's not forget to remove this

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -257,28 +259,28 @@ def fvarNameSlot (fv : Expr) : MetaM (Option Name) := do
...
-  | leaf (fvar : Expr)
-  | pair (left right : FVarTree)
-deriving Inhabited
-
-/-- A tree of binder names, obtained from an `FVarTree`. -/
-inductive NameTree where
-  | leaf (name : Option Name)
-  | pair (left right : NameTree)
+/-- A generic binary tree with data at the leaves. Underlies `FVarTree` and `NameTree`. -/
+inductive BinTree (α : Type) where
```
sonmarcho: What about `BTree`? That's a standard name.

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -560,61 +561,108 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
   Utils.addDeclTac name e (← inferType e) (asLet := false) fun e => do
     trace[Step] "Introduced the \"pretty\" let binding: {← inferType e}"
 
+/-- Recursively destructure an introduced fvar of a Prod type according to the
+shape of a `BinTree`. Returns the leaf fvarsand the updated goal. -/
```
sonmarcho: We should add a not about performance: repeatedly doing `goal.cases` might be expensive - there is probably a way of minimizing the number of operations and the size of the proof term (no need to do anything for now).

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -624,53 +672,77 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
-  /- Introduce the outputs.
+  /- Introduce the single output and recursively destructure it according to the
+     merged binder tree. We use a fresh internal name here and rename leaves to
+     user-provided names later. -/
+  let mut outputFVars : Array FVarId := #[]
+  let goal ← getMainGoal
+  let goalTy ← instantiateMVars (← goal.getType)
+  let goalTy := goalTy.consumeMData
+  if goalTy.isForall then
+    if ¬ (← isProp goalTy.bindingDomain!) then
```
sonmarcho, line 695: I think we need to explain why there might be no quantifier (it happens if the function outputs `()`, in which case we eliminate the `forall` quantifer when doing the simplification above).

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -624,53 +672,77 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
+  /- Total number of "logical" outputs+post slots, used for the user-provided-id check. -/
+  let totalSlots := outputFVars.size + outputIsProp.size
+  if totalSlots < args.ids.size ∧ args.idsUserProvided then
+    logWarning m!"Too many ids provided ({args.ids}): expected ≤ {totalSlots} ids, got {args.ids.size}"
+
+  /- The output count is the number of leaf fvars produced by destructuring. -/
+  let outputCount := outputFVars.size
+  /- The post-condition count: the trailing prop quantifiers/implications. -/
+  let postCount := outputIsProp.size
+  trace[Step] "Output count: {outputCount}, post count: {postCount}"
```
sonmarcho: I prefer `prefixLength`/`postfixLength`: there might be existentially quantified variables (that are not propositions) in the post-condition:
```
f {{ fun x y => A x /\ exists z, P y z }}
```
Could you add a corresponding test so that we make sure this doesn't break in the future (that is: the existentially quantified variables are properly named)?

And we need to update the explanations accordingly (they're deprecated).

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -624,53 +672,77 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
       match args.ids[i] with
       | none => mkFreshAnon isProp
       | some n => pure n
     else
-      -- Otherwise, it depends on whether the var is a prop or not
       if ¬ isProp then
-        -- Generate a name for an output var
         mkFreshUserName `x
       else
-        -- Generate a name for a post-condition
```
sonmarcho, line 674: Why remove the comments?

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -681,29 +753,28 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
-  -- Introduce the pretty equality
-  introPrettyEquality args fExpr (outputs.map Expr.fvar)
+  /- Introduce the pretty equality. -/
+  withMainContext do
+  introPrettyEquality args fExpr (outputFVars.map Expr.fvar)
   traceGoalWithNode "goal after introducing the pretty equality"
 
-  -- Introduce the remaining outputs
+  /- Compute names for the post-conditions. -/
+  let postsIds ← (List.range postCount).mapM fun i => mkFreshName i (outputCount + i) true
```
sonmarcho: Cf remark above: the variable is not necessarily a proposition.

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -560,61 +561,108 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
   Utils.addDeclTac name e (← inferType e) (asLet := false) fun e => do
     trace[Step] "Introduced the \"pretty\" let binding: {← inferType e}"
 
+/-- Recursively destructure an introduced fvar of a Prod type according to the
+shape of a `BinTree`. Returns the leaf fvarsand the updated goal. -/
+partial def destructureFVar {α} (goal : MVarId) (fv : FVarId) (tree : BinTree α) :
```
sonmarcho: I think we should explain (in a docstring) that this function only destructures the fvars, and we rename them later.

sonmarcho (commented):
Thanks for this, it's a massive improvement! I put a few comments here and there.

backends/lean/Aeneas/Std/Primitives.lean
mpenciak: > Is that really general enough? For instance, why don't we need to handle the case `a -> b -> c -> d -> Prop` (we don't want a specific lemma for this of course)?

I think these are all we need. `buildUncurryLam` only ever emits pairs, and if there's multiple top level binders (like `{{(a, b) c d e}}`) then `uncurry'` will bundle them together into one something of the form `a -> b -> Prod _ _ -> Prop` which is handled by `uncurry_eq_prop_arrow`.

backends/lean/Aeneas/Tactic/Step/Tests/IntroOutputs.lean
mpenciak: > Would it be possible to add tests with #guard_msgs showing the environment after the first call to `step`? It would be useful to make sure we get the proper destructuring (here we want to check that `a` and `b` in the spec got substituted with `a.fst` and `a.snd` I think).

I added some trace states. Looking at the proof state it appears like things may be destructuring further than they should? Let me know how it looks, and if it's how you expect things should go.

sonmarcho: I think it looks good: it seems the output is more destructured than it should because your post-condition introduces projection as well. Can you add the following test?

```lean
def pred (_ : Nat × Nat) : Prop := True

@[scoped step]
theorem quadProg_spec :
    quadProg ⦃ a b => pred a /\ pred b ⦄ := by
  unfold quadProg; step*

/- We should get:
a_post1 : pred a
a_post2 : pred (b, c)
-/
example : (do let (a, (b, c)) ← quadProg
  ok (a.1 + a.2 + b.1 + b.2)) ⦃ res => res = 38 ⦄ := by
  step with quadProg_spec
```

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -561,61 +562,113 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
   Utils.addDeclTac name e (← inferType e) (asLet := false) fun e => do
     trace[Step] "Introduced the \"pretty\" let binding: {← inferType e}"
 
+/-- Recursively destructure an introduced fvar of a Prod type according to the
+shape of a `BTree`. Returns the leaf fvarsand the updated goal. The destructured FVars
```
sonmarcho:
```suggestion
shape of a `BTree`. Returns the leaf fvars and the updated goal. The destructured FVars
```

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -561,61 +562,113 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
...
+are named later in `introOutputs`.
+
+NOTE: We are using `cases` to break up the FVar which could be slow for nested
+goals or when the goal context gets large. Something to keep an eye on
```
sonmarcho:
```suggestion
goals or when the goal context gets large. Something to keep an eye on.
```

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -561,61 +562,113 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
...
+      let leftFV := fields[0]!.fvarId!
+      let rightFV := fields[1]!.fvarId!
+      let (lFVs, g1) ← destructureFVar subgoal.mvarId leftFV l
+      let (rFVs, g2) ← destructureFVar g1 rightFV r
+      return (lFVs ++ rFVs, g2)
+    else
+      throwError "destructureFVar: cases on Prod returned {subgoals.size} subgoals, expected 1"
+
+/-- Introduce one universally-quantified output and destructure it according to
+the tree's shape. -/
```
sonmarcho: We should also add a note here about the fact that we will name the fvars later.

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -625,41 +678,71 @@ def introOutputs (args : Args) (fExpr : Expr) (stepState : StepState) :
...
+  if goalTy.isForall then
+    if ¬ (← isProp goalTy.bindingDomain!) then
+      let (fvs, goal') ← introOneSurfaceBinder goal callSiteTree
+      setGoals [goal']
+      outputFVars := fvs
+    else
+      trace[Step] "leading quantifier is a Prop — no output to introduce"
+  else
+    -- There might be no quantifier if the function outputs `()`, in which case we
+    -- eliminate the `forall` quantifier (via `forall_unit`) when doing the simplification above.
```
sonmarcho: I think this comment should be put before the test `if goalTy.isForall` (as it explains why we have to do this sort of checks in the first place).

sonmarcho (commented):
Just a few minor modifications and we're good. Thanks!

mpenciak:
I think the PR is good to go! I added the test and made the adjustments you suggested.

sonmarcho:
Great, thanks! I'm merging.

PR #1054, opened 2026-05-22
fix: `step*?` trace generation in `inferPost`  branches

### LINKED ISSUE #1026: bug: The output of `step*?` is wrong
https://github.com/AeneasVerif/aeneas/blob/e13f77a7804d84b511b897c16a15d9c3f4e54cbf/backends/lean/Aeneas/Tactic/Step/Tests/HigherOrder.lean#L44

mpenciak:
This PR fixes #1026

The core of the issue was that we were calling finalizing tactics right after inferring the postcondition in `stepWith`, without delegating it to `step*`'s `onResult` or `onFinish` (so the tactic wasn't being added to the transcript).

This PR also cleans up the `inferPost` implementation a little bit by pulling the logic out of being inlined in `onBind`. We also fix a small issue where inaccessible names were being introed in the suggestions.


sonmarcho (approved):
That's nice and clean, thanks!

PR #1056, opened 2026-05-25
Improve the performance of the translation

sonmarcho:


sonmarcho:
I merged this PR with https://github.com/AeneasVerif/aeneas/pull/1055 because the latter needed the performance improvement before being merged.

Issue #1027, opened 2026-05-13
feat: Improve the inference of post-condition names for step/step*


sonmarcho:
See for instance: https://github.com/AeneasVerif/aeneas/blob/e13f77a7804d84b511b897c16a15d9c3f4e54cbf/backends/lean/Aeneas/Tactic/Step/Tests/TupleDestruct.lean#L45
We get:`let* ⟨ a, b, a_post1, a_post2 ⟩ ← twoNats_spec` while something like `let* ⟨ a, b, a_post, b_post ⟩ ← twoNats_spec` would be more meaningful.

sonmarcho:
Fixed by https://github.com/AeneasVerif/aeneas/pull/1058

PR #1058, opened 2026-05-25
feat: Improve postcondition naming

mpenciak:
This PR addresses #1027.

We now inspect the type of the postcondition, and if it's a binary relation (like `<`, `<=`, `>`, `>=`, `=`, or `==`) with a single variable on one side then we prefer using a name of the orm `<id>_post<index>`. Otherwise we fall back to the previous behavior.


backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -747,56 +772,87 @@ def introOutputs (info : SpecInfo) (args : Args) (fExpr : Expr) (stepState : Ste
...
+      m.insert outputFVars[i]! outputIds[i]!
+
+  /- Now compute the prop-status and the names of the remaining leading binders (these are
+     the post-conditions and existential variables to introduce).
+
+     We prefer user-provided names, and for non-Props we generate fresh names. For Props we
+     analyze the type of the hypothesis: if it is of the form `<id> <binrel> _` then we
+     prefer using `<id>_post<idx>`, so that each post-condition is named after the output
+     it constrains. We compute the names here, while the binders' types are still at hand,
+     and use them further below when actually introducing the binders. -/
```
sonmarcho, line 813:
```suggestion
     We prefer user-provided names, and for non-Props we generate fresh names. For Props we
     analyze the type of the hypothesis: if it is of the form `<id> <binrel> _` (or `_<binrel><id>`)
     then we prefer using `<id>_post<idx>`, so that each post-condition is named after the output
     it constrains. We compute the names here, while the binders' types are still at hand,
     and use them further below when actually introducing the binders. -/
```

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -747,56 +772,87 @@ def introOutputs (info : SpecInfo) (args : Args) (fExpr : Expr) (stepState : Ste
...
+            | some (nameRoot, baseName) =>
+              let suffixNum := postCounts.getD baseName 0
+              let suffixStr := if 0 < suffixNum then s!"{suffixNum}" else ""
+              postCounts := postCounts.insert baseName (suffixNum + 1)
+              pure (Name.str nameRoot s!"{baseName}_post{suffixStr}")
+        postsIds := postsIds.push n
+      pure (typesAndProps.map (fun (p : Expr × Bool) => p.2), postsIds)
+
+  /- The postfix length: the remaining binders (post-condition and existential variables). -/
+  let postfixLength := outputIsProp.size
```
sonmarcho: The name `outputIsProp` doesn't make sense here

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -747,56 +772,87 @@ def introOutputs (info : SpecInfo) (args : Args) (fExpr : Expr) (stepState : Ste
...
+              match ← decomposeRelation? ty outputNames, args.postsBasename with
+              | some relName, some (.str root _) => some (root, relName)
+              | some relName, _                  => some (.anonymous, relName)
+              | none,         some (.str root s) => some (root, Name.mkSimple s)
+              | none,         _                  => none
+            match nameSpec with
+            | none => mkFreshAnonPropUserName
+            | some (nameRoot, baseName) =>
+              let suffixNum := postCounts.getD baseName 0
+              let suffixStr := if 0 < suffixNum then s!"{suffixNum}" else ""
```
sonmarcho:
```suggestion
              let suffixStr := if suffixNum = 0 then "" else s!"{suffixNum}"
```

backends/lean/Aeneas/Tactic/Step/Step.lean
mpenciak: I renamed it to `postsIsProp`, closer to the `postsIdsArr` it mirrors.

sonmarcho (approved):

Issue #1053, opened 2026-05-22
`Iterator::zip` test check fails with `RErased`


waamm:
In an attempt to get more acquainted with `aeneas`, I've been trying to get the following test to work:

```
    let a = [1u32, 2u32, 3u32];
    let b = [9u32, 17u32, 10u32];
    let mut sum = 0u32;
    for (x, y) in a.iter().zip(b.iter()) {
        sum = sum.wrapping_add(*x).wrapping_add(*y);
    }
    assert!(sum == 42);
```

I've tried to add Lean support for `Zip`, `Iterator::zip`, `Zip::next`, etc., but I'm struggling with the symbolic borrow checker, in particular [this](https://github.com/AeneasVerif/aeneas/blob/2620c238c3bf20464920704ef56603f7c441f666/src/interp/InterpBorrowsCore.ml#L408) step. The type of a projector might look like `Zip<Iter<'4, u32>, Iter<'5, u32>> ` so `normalize_proj_ty` turns that into `Zip<Iter<'0>, Iter<'_>>`, i.e. the other array’s lifetime becomes `RErased`.

I'm not sure what the appropriate fix is here; one way of fixing this might be to propagate `allow_erased` from `projection_contains` to `compare_rtys`?

sonmarcho:
I actually had a similar issue with `zip` a few days ago and am working on a fix: we indeed have to propagate `allow_erased` (there are other issues, also).

waamm:
This MWE still fails for me:

```
pub fn arrays_zip_iter() {
    let a = [1u32, 2u32, 3u32];
    let b = [9u32, 17u32, 10u32];
    for _ in a.iter().zip(b.iter()) {}
}
```

I think it's for two reasons:

(i)
The lookup visitor calls `super#visit_aproj` both inside the `AProjBorrows` arm and again unconditionally after the match (same for `visit_eproj`) [here](https://github.com/AeneasVerif/aeneas/blob/21bfef95cd5f09837f4d9549fc30e09afd4951da/src/interp/InterpBorrowsCore.ml#L1552). That double descent can visit the same `AProjLoans` twice and trip the `set_found` sanity check [here](https://github.com/AeneasVerif/aeneas/blob/21bfef95cd5f09837f4d9549fc30e09afd4951da/src/interp/InterpBorrowsCore.ml#L1527).

(ii)
[...]

PR #1065, opened 2026-05-27
Fix double visit in lookup_aproj_loans_opt

waamm:
See issue [1053](https://github.com/AeneasVerif/aeneas/issues/1053).


sonmarcho:
Thanks for this! I need to add a test corresponding to issue https://github.com/AeneasVerif/aeneas/issues/1053.
Also, I'm a bit surprised by the fix (or rather: I'm surprised we need this fix at all): let me investigate a bit.

waamm:
> Thanks for this! I need to add a test corresponding to issue #1053. Also, I'm a bit surprised by the fix (or rather: I'm surprised we need this fix at all): let me investigate a bit.

Just in case you missed it when you changed the PR description: I intended this code change to be part of the solution for #1053, but it doesn't fully fix the issue yet. I'll make a quick commit adding the test and a minimal fix (just 3 more lines) for you to review.

sonmarcho:
Sorry about that: indeed I missed it (I reverted)

PR #1066, opened 2026-05-27
String: prove `toStr` length bound via `ByteArray.length_toList`

remix7531:
The length-bound obligation in `Aeneas.Std.toStr` was `sorry`'d because no lemma relating `ByteArray.toList.length` to `ByteArray.size` was available. Add `ByteArray.length_toList` (proved by induction on `ByteArray.toList.loop`) and use it to discharge the obligation.

No AI was used.


sonmarcho (approved):
Thanks for this!

Issue #828, opened 2026-03-11
Error: unsupported floats


TitiSkywalker:
Hello! I am trying to convert the following Rust code into Lean:
```rust
#[derive(Clone, Debug, PartialEq)]
pub enum BucketReduction {
    /// There are not enough tokens to complete the operation.
    Failure,
    /// A part of the available tokens have been consumed.
    Success,
    /// A number of tokens `inner` times larger than the bucket size have been consumed.
    OverConsumption(f64),
}

fn test_create() -> BucketReduction {
    BucketReduction::OverConsumption(1.5)
}
```

But I get the following error outputs from Aeneas:
``` 
Applied prepasses:  [############################################################] 24/24 ✔️
Translated globals:  [#############################################################] 1/1 ✔️
Translated opaque functions:  [##################################################] 13/13 ✔️
[Error] Improperly typed constant value--------------------------------------------] 0/6 ⠋
Source: 'src/main.rs', lines 12:4-12:41
[Warn ] Could not translate the body of function 'rust_mode::test_create because of previous error
Name pattern: 'rust_mode::test_create'
Definition span: 'src/main.rs', lines 11:0-13:1
Source: 'src/main.rs', lines 12:4-12:41
Compiler source: interp/Interp.ml, line 555
```

Moreover, if I tweak the code a little bit:
```rust
fn test_create(x: f64, y: f64) -> BucketReduction {
    BucketReduction::OverConsumption(x / y)
}
```

The error outputs become:
```
[Error] Invalid inputs for binop:
```

Does anyone know how to fix this problem? :raised_hands:

PS: the code comes form Amazon firecracker. It would be meaningful to be able to verify real-life codes :grinning: !

PR #1073, opened 2026-05-28
Initial draft of basic float support for lean backend

ezrosent:
Initial stab at #828.

All of the code was written by codex, but the current diff is the result of several passes of code review, reading the code, asking clarifications, and asking the agents to restructure the OCaml and make it more readable.

Basic approach is to minimally lower f32 and f64 to Float and Float32 in Lean. Lean's own support for floats does not seem super advanced, so lots of core routines are `opaque`.


jcdong98:
FYI: In Lean 4 (as of v4.33), `Float` and `Float32` are no longer opaque and are now backed by `Float.Model` (see https://github.com/leanprover/lean4/pull/14091).

Once Aeneas bumps its lean-toolchain (currently at v4.31.0) to v4.33+, this could help avoid defining many of the core routines as opaque and enable actual proofs over float operations. Hope this is helpful!

PR #1078, opened 2026-05-29
Add a catch-all branch when matching on statements

Nadrieril:
This makes it possible to add statements in Charon without breaking Aeneas.

### DIFF
--- a/src/PrePasses.ml
+++ b/src/PrePasses.ml
@@ -749,8 +749,9 @@ let remove_useless_joins (crate : crate) (f : fun_decl) : fun_decl =
                 (can_inline, st :: ls)
             | BinaryOp _ | UnaryOp _ | Discriminant _ | Len _ | Repeat _ ->
                 (false, st :: ls))
-        | SetDiscriminant _ | CopyNonOverlapping _ | Assert _ | Call _ | Error _
-          -> (false, st :: ls))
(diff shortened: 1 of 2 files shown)


src/PrePasses.ml
```diff
@@ -749,8 +749,9 @@ let remove_useless_joins (crate : crate) (f : fun_decl) : fun_decl =
                 (can_inline, st :: ls)
             | BinaryOp _ | UnaryOp _ | Discriminant _ | Len _ | Repeat _ ->
                 (false, st :: ls))
-        | SetDiscriminant _ | CopyNonOverlapping _ | Assert _ | Call _ | Error _
-          -> (false, st :: ls))
+        | SetDiscriminant _ | Assert _ | Call _ | Error _ -> (false, st :: ls)
+        | _ -> failwith ("unsupported statement: " ^ show_statement_kind st.kind)
```
sonmarcho, line 753: I just noticed that: we want to use `[%craise]` instead of `failwith` to have a recoverable failure

src/PrePasses.ml
```diff
@@ -1593,6 +1591,9 @@ let decompose_global_accesses (crate : crate) (f : fun_decl) : fun_decl =
             | Switch _
             | Loop _
             | Error _ -> st.kind
+            | _ ->
+                failwith
+                  ("unsupported statement: " ^ show_statement_kind st.kind)
```
sonmarcho, line 1596: Similar issue

PR #1092, opened 2026-06-01
Fix minor issues

sonmarcho:
Fixes minor issues introduces by https://github.com/AeneasVerif/aeneas/pull/1090


Nadrieril:
Makes sense, but you already have a check that errors on GATs at definition site, no?

sonmarcho:
Yes, but once I start adding support for GATs it's useful to have sanity checks for all the places that need to be updated - I like being defensive in Aeneas' code.

Nadrieril:
Good point, makes sense

PR #1114, opened 2026-06-10
Prepare for the removal of method declaration items

Nadrieril:
This PR prepares Aeneas for the removal in Charon of the fake `FunDecl`s we add for each method declaration (https://github.com/AeneasVerif/charon/issues/1148). They are misleading because methods are more often an abstraction boundary than a concrete item, and in particular it's useful to distinguish "the method declaration" from "the method default implementation" (when there is one). Using the same function item for both creates confusion.

The main change is that a couple of maps that used `FunDeclId`s as keys now use `FunOrMethodId`.

### DIFF
--- a/src/interp/InterpStatements.ml
+++ b/src/interp/InterpStatements.ml
@@ -579,9 +579,8 @@ let mk_symbolic_fun_call_inst (span : Meta.span) (ctx : eval_ctx)
       instantiate_fun_sig (Some span) ctx inst_generics tr_self signature;
   }
 
-(** Auxiliary helper for [eval_non_builtin_function_call_symbolic] Instantiate
-    the signature and introduce fresh abstractions and region ids while doing
-    so.
+(** Auxiliary helper for [eval_function_call_symbolic] Instantiate the signature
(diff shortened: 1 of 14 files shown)


src/interp/InterpStatements.ml
```diff
@@ -1358,15 +1389,22 @@ and eval_function_call_concrete (config : config) (span : Meta.span)
...
+  | FnOpRegular func ->
+      [%ltrace
+        "function call:\n- call: " ^ call_to_string ctx call ^ "\n- fn_ptr : "
+        ^ fn_ptr_to_string ctx func];
+      let call_info = eval_function_call_symbolic_inst span func ctx in
+      [%ltrace
+        "- call: " ^ call_to_string ctx call ^ "\n- call.generics:\n"
+        ^ generic_args_to_string ctx call_info.call_generics
+        ^ "\n- def.signature:\n"
+        ^ fun_sig_to_string ctx call_info.call_sig];
```
Nadrieril, line 1404: There is a functional change in tracing: now we get that info for all function calls instead of only some. That should be the only functional change.

PR #1120, opened 2026-06-11
Deduplication of identifiers arising from traits

### LINKED ISSUE #1051: Duplicate name when two supertraits are the same trait
**See the comment below for a better example of this same problem.** 

When Aeneas extracts code that uses `std::num::NonZero*` types, the generated Lean for the `core::num::nonzero::ZeroablePrimitive` trait declaration contains a **duplicate field name** in both the `parentClauses` attribute and the structure body.

`ZeroablePrimitive` has three super-traits:
1. `core::marker::Copy for Self`
2. `core::num::nonzero::private::Sealed for Self`
3. `core::marker::Copy for Self_NonZeroInner`

Traits 1 and 3 are the *same* trait (`Copy`) for *different* type arguments. Aeneas generates the same name (`markerCopyInst`) for both, producing a duplicate field name.

## Minimal Rust example

```rust
use std::num::NonZeroU8;

pub fn get_inner(x: NonZeroU8) -> u8 {
    x.get()
}
```

oliver-butterley:
**Draft PR: I'm still looking at the details of this.**

A trait with two associated types which require the *same* trait produces two parent clauses which instantiate that trait with different type arguments. E.g.,

```rust
pub trait KeySerde {
    fn serialize(&self) -> u32;
}

pub trait KeyPairSerde {
    type PublicKey: KeySerde;
    type PrivateKey: KeySerde;
    fn get_public(&self) -> Self::PublicKey;
    fn get_private(&self) -> Self::PrivateKey;
}
```

Before the fix, Aeneas generated the same field name (`KeySerdeInst`) for the fields. This fix improves the naming algorithm . If there is no duplication then naming is as before, if there is duplication then a suffix is added. For the example given, the two fields become:

```lean
  KeySerdeSelf_PublicKeyInst : KeySerde Self_PublicKey
  KeySerdeSelf_PrivateKeyInst : KeySerde Self_PrivateKey
```

AI usage: AI used but all code carefully checked by human and modified.

fixes #1051


oliver-butterley:
Closing this in favour of #1142. In this PR the implementation of the improved functionality is done with a minimal amount of changed code but in a way that the functionality is spread across many files in a way that isn't really maintainable.

PR #1128, opened 2026-06-16
De-duplicate parent trait names

mpenciak:
This PR addresses #1051.

If there are no potential duplicates, the original `<traitname>Inst` is used, but if there are two parents which implement the same trait the new deduplicated name is of the form `<traitname><parent>Inst`. Adds a regression test to demonstrate the fix.


mpenciak:
I just realized #1120 already fixes this issue, and the solution seems more robust so I'd be happy to close this one if we go with that fix instead.

oliver-butterley:
> I just realized #1120 already fixes this issue, and the solution seems more robust so I'd be happy to close this one if we go with that fix instead.

As per the other conversation, I think this PR is stronger that mine. I particularly like how all the deduplication logic is isolated in one place. I strongly suggest going for this PR.

oliver-butterley:
A few comments.

-------

Would it be possible to add a test that triggers the case of adding a numerical suffix? If I understood correctly, this would be something like
```rust
// Test for numeric fallback in deduplication: two associated types both bounded by 
// the same trait with no type parameters.
pub trait PairTrait {
    type First: Clone;
    type Second: Clone;
    fn first(&self) -> &Self::First;
    fn second(&self) -> &Self::Second;
}

pub fn clone_pair<T: PairTrait>(x: &T) -> (T::First, T::Second) {
    (x.first().clone(), x.second().clone())
}
```

------

Currently the code 
- adds the `Inst` suffix in [extractBase.ml:2006](https://github.com/mpenciak/aeneas/blob/bb46147ca452816786b134e82e8216bf4fcb5e7e/src/extract/ExtractBase.ml#L2006),
- removes the `Inst` suffix in [extract.m:2710](https://github.com/mpenciak/aeneas/blob/bb46147ca452816786b134e82e8216bf4fcb5e7e/src/extract/Extract.ml#L2710)
- adds the `Inst` suffix again after possibly adding something in the middle in [extract.ml:2729](https://github.com/mpenciak/aeneas/blob/bb46147ca452816786b134e82e8216bf4fcb5e7e/src/extract/Extract.ml#L2729).

This feels a bit wrong although I can't see immediate problems. I could play at creating Rust code using Inst in the names... :rofl: Anyway, is it possible to make this rather more solid? However this really depends on a complete overview of the systems Aeneas uses to assign names and I'm not very aware of all the subtle details.

oliver-butterley:
@sonmarcho, could you give some thoughts about the general architecture of the naming code? 

The way it is arranged at the moment there are multiple functions defined in `ExtractBase.ml`, like `ctx_compute_trait_parent_clause_name` (the one relevant to this particular case). These functions are the ones which create the name for an individual object. We can't add the new feature by updating this because it doesn't have knowledge of potential siblings and so we do a later pass. Another option would be to pass to this function all the info about everything else. Then it could check if there would be any name collisions and assign names accordingly.

mpenciak:
Closing this in favor of #1142, which looks like it's in a good state now.

PR #1133, opened 2026-06-17
Feat/range iterator next step spec

### LINKED ISSUE #1132: core::iter::range::IteratorRange::next has no reasoning lemma, range-for loops can't be proven
The range iterators in `backends/lean/Aeneas/Std/Core/Iter.lean` are extracted as plain `def`s with no `@[step]` / `@[progress]` / `@[spec]` / `@[simp]` lemma attached. So `step` (and `progress` / `let*` in the older era) cannot step through the body of a `for i in a..b` loop, because there's nothing to apply at the `next` call. Every downstream project that wants functional correctness through a range loop re-derives a `next` lemma by hand.

### Repro

Take any range-`for` loop reduced to the `loop` + `next` shape and try to drive it with `step`:

```lean
example (r : core.ops.range.Range Usize) :
    loop (fun iter => do
      let (o, iter') ← core.iter.range.IteratorRange.next core.iter.range.StepUsize iter
      match o with
      | none => ok (ControlFlow.done ())
      | some _ => ok (ControlFlow.cont iter')) r
    ⦃ (_ : Unit) => True ⦄ := by
  apply loop.spec_decr_nat (measure := fun s => s.«end».val - s.start.val) (inv := fun _ => True)
  · intro s _
    step      -- "could not find a local assumption or a theorem to apply"
    ...
```

`step` has nothing to fire on the `next` call.

### What I checked

- No lemma about `IteratorRange.next` exists anywhere in `backends/lean/` or `tests/lean/`.
- The 9 test files that extract range-`for` loops prove zero of them. Functional correctness through the range iterator is unexercised in-tree.
- The `proof-patterns` skill shows a range-loop template that reads as if `step*` already handles this. It doesn't today: un-wire a `next` spec and that template stops compiling. The doc is ahead of the library.

This isn't a soundness issue and it isn't a bug in extraction (range loops extract fine). It's a missing library lemma: Aeneas can extract range loops but can't ergonomically prove them.

### Fix

I have a PR coming right behind this. It adds a single deterministic `@[step]` WP spec for the `usize` range iterator's `next` (if `start < end` it yields `start` and advances by one, else it reports exhaustion), wired into `Std/Core.lean`, plus a regression test under `Tactic/Step/Tests/` that proves the loop above is total with zero hand-written glue. Builds clean against current `main` (full lib and `tests/lean` both green).

One scoping note for the review: I specced `StepUsize` only, since that's the only `Step` instance instantiated today. If you'd rather have the spec stated generically over `Step` (or want me to update the `proof-patterns` doc to match), say so on the PR and I'll adjust.

eslerm:
Closes #1132.

The `usize` range iterator's `next` ships with no reasoning lemma, so `step`
can't drive a `for i in a..b` loop. This adds a deterministic `@[step]` WP spec
for it (`Std/Core/IterSpec.lean`, wired into `Std/Core`), plus a regression test
proving a range-`for` loop total via the spec (`Tactic/Step/Tests/RangeLoop.lean`).
Full lib and `tests/lean` build green; `#print axioms` is the three standard ones.

I specced `StepUsize` only, the one `Step` instance today. Happy to generalize
over `Step` if you'd prefer.

### DIFF
--- a/backends/lean/Aeneas/Std/Core.lean
+++ b/backends/lean/Aeneas/Std/Core.lean
@@ -10,6 +10,7 @@ import Aeneas.Std.Core.Fmt
 import Aeneas.Std.Core.FmtWithSlice
 import Aeneas.Std.Core.Hash
 import Aeneas.Std.Core.Iter
+import Aeneas.Std.Core.IterSpec
 import Aeneas.Std.Core.Marker
 import Aeneas.Std.Core.Ops
 import Aeneas.Std.Core.Panic
--- a/backends/lean/Aeneas/Std/Core/IterSpec.lean
+++ b/backends/lean/Aeneas/Std/Core/IterSpec.lean
@@ -1,0 +1,64 @@
+/- Specs for the `core::iter` iterators.
+
+   The iterator `next` functions (Core/Iter.lean) are extracted as plain `def`s
+   with no reasoning lemmas, so `progress` / `let*` cannot step through a
+   `for i in a..b` loop: every downstream project re-derives a `next` lemma by
+   hand. This file provides the missing `@[step]` spec for the `usize` range
+   iterator, the overwhelmingly common case. -/
+import Aeneas.Std.Core.Iter
+import Aeneas.Tactic.Step.Init
+
+namespace Aeneas.Std
+
+open Result Error WP
+
+/-- **Spec for the `usize` range iterator's `next`** (the engine of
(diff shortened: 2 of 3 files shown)


eslerm:
**AI Assistance**: this PR sits somewhere between vibe coded and agentically engineered. I came across this bug while adding constant time bignum to https://github.com/hexproofdev/coprime One can scaffold around this bug, but a common upstream fix seemed worth suggesting.

PR #1142, opened 2026-06-20
Deduplication of identifiers arising from traits

### LINKED ISSUE #1051: Duplicate name when two supertraits are the same trait

### LINKED ISSUE #1245: Bug: Lean backend emits duplicate parent-clause field names

```rust
pub trait Pair {
    type Left: Copy;
    type Right: Copy;
}

pub struct PairImpl;

impl Pair for PairImpl {
    type Left = bool;
    type Right = u8;
}
```

The crate uses Rust edition 2024 and has no dependencies.

**Charon command**:

```text
charon cargo --preset=aeneas --dest-file=repro.llbc
```

**Aeneas command**:

```text
aeneas -backend lean -dest lean repro.llbc
```

The generated file was checked with:

```text
lake env lean lean/Repro.lean
```

**Versions**:
- Aeneas: `daa85d7e89400fa978be83fedbc7e475a83f0889`
- Charon: `0.1.232` (`340b1af4df92608d0911fc2ba26eef3fd3a30ab4`)
- Lean: `4.32.1`

**Aeneas output**:

Aeneas exits successfully and reports:

```text
[Info] Imported: repro.llbc
[Info] Generated: lean/Repro.lean
[Info] Total execution time: 0.099954 seconds
```

Lean rejects the generated file:

```text
lean/Repro.lean:22:2: error: Field `coremarkerCopyInst` has already been declared
lean/Repro.lean:33:60: error: Function expected at
  Pair
but this term has type
  ?m.1
```

**Lean translation**:

Aeneas generates the same field name for both parent clauses:

```lean
structure Pair (Self : Type) (Self_Left : Type) (Self_Right : Type) where
  coremarkerCopyInst : core.marker.Copy Self_Left
  coremarkerCopyInst : core.marker.Copy Self_Right
```

The implementation literal repeats the field name as well:

```lean
{
  coremarkerCopyInst := core.core.marker.CopyBool
  coremarkerCopyInst := core.marker.CopyU8
}
```

The generated fields need unique names. Including the associated type or a stable clause index would both produce valid Lean, for example `selfLeftCoremarkerCopyInst` and `selfRightCoremarkerCopyInst`.

**Other steps needed to reproduce**:

The generated file imports the Aeneas Lean package at the same commit. No handwritten Lean is required.

`--remove-unused-clauses` can erase unused marker bounds in examples where they are unnecessary. Post-processing the generated file also works. Neither changes the fact that the default Lean backend emits syntactically invalid code when same-named parent clauses remain.

**Explain the bug**:

Parent-clause field names are derived from the bound trait alone. Two associated types carrying the same bound therefore collide in the generated Lean structure and every corresponding implementation literal. Aeneas reports successful extraction even though Lean cannot elaborate the artifact.

**Priority / impact**:

This blocks direct refinement proofs for generic protocol traits with equally constrained associated types. Concrete free-function extraction still works, but requires avoiding the production trait or mechanically editing generated output.

**Where was this seen?**:

Reduced from a production state-machine trait used by Dedalus Machines. Its phase and observation associated types both carry `Copy`, `PartialEq`, `Eq`, and `Debug` bounds, causing four duplicate field names.

**Regression?**:

Unknown.

oliver-butterley:
Previously the scheme failed to allocate unique names in the case when a trait has two associated types, both bounded by the same trait.

This PR updates the naming scheme. When there is no collision then multiple parts are dropped, they are only included and the names become longer when required to prevent collisions.

A new test file is added with several examples to test various related edge cases. 

As a by product of the updated naming scheme, the duplication previously seen in the coq output is also now eliminated. 

+~300 due to the new test, core logic change is +~130 / -~90

*This solves the same problem as #1128 and #1120. Multiple draft PRs is due to there being several distinct strategies to check and then compare.*

Fixes #1051 
Fixes #1245

Co-authored by @mpenciak


oliver-butterley:
The naming algorithm is being updated to handle the problem when a specific combination of names causes a collision. I.e.,
- `Parent`+`Self`+`Foo`+`SideBar` = `ParentSelfFooSideBar`
- `Parent`+`Self`+`FooSide`+`Bar` = `ParentSelfFooSideBar`
For these unlikely cases we add a numerical suffix.

src/extract/Extract.ml
```diff
@@ -2686,39 +2686,16 @@ let extract_global_decl (ctx : extraction_ctx) (fmt : F.formatter)
...
-          = List.length info.parent_clauses)
-          ("Invalid builtin information for trait decl: "
-          ^ name_to_string ctx trait_decl.item_meta.name
-          ^ "; expected "
-          ^ string_of_int (List.length trait_decl.parent_clauses)
-          ^ " parent clauses, found "
-          ^ string_of_int (List.length info.parent_clauses));
-        List.map
-          (fun (c, name) -> (c.clause_id, name))
-          (List.combine trait_decl.parent_clauses info.parent_clauses)
```
oliver-butterley, line 2730: This functionality is now contained within ExtractBase.ml, essentially the same code as here.

src/extract/ExtractBase.ml
```diff
@@ -341,13 +341,13 @@ let unsafe_names_map_add (id : id) (name : string) (nm : unsafe_names_map) :
...
-    (append : string -> int -> string) (basename : string) : string =
+let string_to_unique (collision : string -> bool)
+    (append : string -> int -> string) (name : string) : string =
   let rec gen (i : int) : string =
-    let s = append basename i in
+    let s = append name i in
     if collision s then gen (i + 1) else s
   in
-  if collision basename then gen 1 else basename
+  if collision name then gen 1 else name
```
oliver-butterley: There is no practical need to change the name of the function from `basename_to_unique_aux` to `string_to_unique`. However we now use this function in another place, not just for basenames and so the naming was confusing when it was used. The new name describes what it actually does.

tests/src/issue-1051-duplicate-parent-clause-names.rs
```diff
@@ -0,0 +1,55 @@
...
+pub trait TraitE {
+    type X: TraitA;
+    type Y: TraitC;
+}
+
+// Example 3: This function pulls in `NonZeroU8` and hence `ZeroablePrimitive`. This has two `Copy`
+// parent clauses which would collide on the same field name without deduplication.
+pub fn get_inner(x: NonZeroU8) -> u8 {
+    x.get()
+}
```
oliver-butterley, line 28: I was unsure about including this example since it doesn't test something related to this issue that the previous examples don't already test. However it does it in a way where the naming is particularly confused so I thought it best to include to have more examples to avoid errors with naming in the future.

tests/src/issue-1051-duplicate-parent-clause-names.rs
```diff
@@ -0,0 +1,55 @@
...
+
+// Example 5: Two parent clauses of the same generic trait, they differ in a const-generic argument.
+// These don't need deduplication since the base names shouldn't collide.
+pub trait GenN<const N: usize> {}
+
+pub trait UsesGenN: GenN<3> + GenN<5> {}
+
+// Example 6: The specific combination of names causes a collision. I.e.,
+// `Parent`+`Self`+`Foo`+`SideBar` = `ParentSelfFooSideBar`
+// `Parent`+`Self`+`FooSide`+`Bar` = `ParentSelfFooSideBar`
```
oliver-butterley, line 46: This example truly adds a very different problem compared to the first ones. It is due to not using separators. I expect it doesn't happen very often in practice but not impossible.

src/extract/ExtractBase.ml
```diff
@@ -1990,27 +1990,137 @@ let ctx_compute_trait_clause_name (ctx : extraction_ctx)
...
-     For instance, if the clause gives us an instance of `Foo<u32>`,
-     we generate a name along the lines of "fooU32Inst".
-  *)
-  (* We need to lookup the LLBC definitions, to have the original instantiation *)
-  let clause =
-    let current_def_name = trait_decl.item_meta.name in
-    let params = trait_decl.llbc_generics in
-    ctx_compute_trait_clause_name ctx current_def_name params
-      trait_decl.llbc_parent_clauses clause.clause_id
+(** Given a type argument, calculate a discriminator fragment in PascalCase. *)
```
sonmarcho: That's Camel case actually

src/extract/ExtractBase.ml
sonmarcho: I'm going to rename it to `name_to_unique`

src/extract/ExtractBase.ml
oliver-butterley: Indeed... oops!

oliver-butterley:
**Further investigation in progress**. We shouldn't need to make so many modifications to the code: the name generation the way it is should already give us unique names (because it already derives the names from the trait clauses). Perhaps the problem we have here is that the name generation automatically filters variable names and so we should add an option to filter them when generating names for top-level impls, but not filter them when generating names for structure fields that model super traits.

mpenciak (commented):
Only comment is that you need tho wire the the new test into `tests/lean/lakefile.lean`, otherwise looks good! A nice improvement on both #1128 and #1120.

oliver-butterley:
Thanks for taking a look, much appreciated! 

> Only comment is that you need tho wire the the new test into `tests/lean/lakefile.lean`, otherwise looks good! 

Good catch with the lakefile, there is a story behind this... :rofl: I did update it using the script (`tests/lean/script.sh`) which gets called by make. However the script doesn't enforce the sort order. So I undid that, and then forgot to manually add the new entry. I've now added it.

On this machine I get:
```
AdtBorrows < Adt < ArraysDefs < Arrays < ...
```
Whereas commited is:
```
Adt < AdtBorrows < Arrays < ArraysDefs < ...
```
Would be convenient to fix the sort order in the script! So another little PR... :wink: @sonmarcho  EDIT: This is PR https://github.com/AeneasVerif/aeneas/pull/1263

oliver-butterley:
Another option is revert the changes to the ocaml code in this PR and instead to add:
```ocaml
  let _, clause_names =
    List.fold_left
      (fun (used, clause_names) (cid, cname) ->
        let cname =
          basename_to_unique_aux
            (fun candidate -> StringSet.mem candidate used)
            name_append_index cname
        in
        (StringSet.add cname used, (cid, cname) :: clause_names))
      (StringSet.empty, []) clause_names
  in
  let clause_names = List.rev clause_names in
```
in the definition of `extract_trait_decl_register_parent_clause_names` in `src/extract/Extract.ml`. (Fully implemented in [this branch](https://github.com/oliver-butterley/aeneas/tree/1142))

This also makes sure that all the examples in the test are given unique names. However more of the examples become unique simply by having a numeric suffix added rather than meaningful names for deduplication. 

@sonmarcho we'll need to opinion on the design choice here, how important are meaningful names for deduplication in this context?

PR #967, opened 2026-05-05
Add bitwise operators on booleans.

maximebuyse:
Fixes https://github.com/AeneasVerif/aeneas/issues/965

Adds support for bitwise operators on booleans (`&`, `|`, `^`). Their semantics is different than regular boolean `&&` and `||` as they are not short-circuiting. This is why we need to introduce some nodes in the pure AST for `BoolAnd` and `BoolXor` (`BoolOr` already exists as it can be produced by the translation).

Issue #965, opened 2026-05-04
`&` on `bool` gives `Error: Invalid inputs for binop`


maximebuyse:
## Rust reproducer

```rust
pub fn check(a: bool, b: bool) -> bool {
    a & b
}
```

maximebuyse:
Same issue for `BitOr` (`a | b`) and `BitXor` (`a ^ b`) on bool

niooss-ledger:
Hello,
I stumbled on this issue while analyzing function `xor_long` from https://github.com/ayhon/sha3.rs/blob/f186e6fbd2a1424617166efe11d3daa682583324/src/simple.rs#L186-L190

```rust
fn xor_long(s: &mut [bool], other: &[bool]){
    for (si, oi) in s.iter_mut().zip(other) {
        *si ^= *oi;
    }
}
```

This makes Aeneas report "Error: Invalid inputs for binop" because `*si ^= *oi` operates on booleans and I confirmed Pull Request https://github.com/AeneasVerif/aeneas/pull/967 fixed the issue.

By the way, it seems https://github.com/AeneasVerif/sha3.rs introduced an extra function `binxor` to work around this issue (https://github.com/AeneasVerif/sha3.rs/blob/cb411d54ee0b22f88b64c79963691ff63f3ad5b4/src/simple.rs#L8-L20):

```rust
fn binxor(a: bool, b: bool) -> bool {
    (a && !b) || (b && !a)
}

fn xor_long_at(s: &mut [bool], other: &[bool], pos: usize){
    // TODO: Remove once fixed in Aeneas
    let n = (s.len()).min(other.len() + pos);
    let mut i = pos;
    while i < n {
        s[i] = binxor(s[i], other[i-pos]);
        i += 1;
    }
}
```

(N.B. I don't know whether the `TODO` comment refers to this issue (https://github.com/AeneasVerif/aeneas/issues/965) or to the fact that `zip` is not modeled (https://github.com/AeneasVerif/aeneas/issues/1053).)

Anyway, it would be useful to get binop on bools properly supported in Aeneas and https://github.com/AeneasVerif/aeneas/pull/967 seems to be achieve this in a quite-straightforward way. Could this be reviewed?

oliver-butterley:
[...]

PR #1168, opened 2026-06-25
feat: Add support for boolean comparisons

sonmarcho:
This PR adds support for `<`, `<=`, `>`, `>=` applied to booleans.


src/interp/InterpExpressions.ml
```diff
@@ -1106,8 +1124,13 @@ let eval_binary_op_symbolic (config : config) (span : Meta.span) (binop : binop)
         "The type is not primitively copyable";
       TLiteral TBool)
     else
-      (* Other operations: input types are integers *)
+      (* Other operations: input types are integers, with the exception of the
+         comparison operations which also accept booleans (booleans are ordered:
+         [false < true]). *)
```
niooss-ledger, line 1129: @sonmarcho This comment/implementation is incomplete: bitwise operations also accept booleans. This was reported in issue  #965 and PR #967, which seem to be ignored for months. What would be the best way to make you aware of such similar issues?

sonmarcho: Sorry about that. The issue is that we're lacking ressources to properly triage the issues and the PRs, with the consequence that we regularly miss or forget things. @oliver-butterley as you want to help with triaging let's discuss this during our next meeting. @niooss-ledger are you on the Zulip? It will be easier to coordinate there.

oliver-butterley: Absolutely @sonmarcho! Moreover bitwise operations are already on my high priority list for our next meeting. 

@niooss-ledger I believe things are already improving with the immense backlog and soon will be much better. Let's keep in touch, either here or in one of the Zulips.

niooss-ledger: Nice to hear! On https://aeneas-verif.zulipchat.com/ my user name is IooNagLdg

PR #1169, opened 2026-06-25
Avoid reserved keywords being used as a bare name

### LINKED ISSUE #1105: Bug: `name` is a keyword in the Lean backend
It seems some syntax in the Lean backend turned `name` into a keyword, meaning we can't use it as an identifier anymore.

### LINKED ISSUE #1163: Lean keywords used as function names aren't escaped
E.g., 
```rust
pub fn example(x: u32) -> u32 {
    x
}
```

This function is translated by Aeneas to 

```lean4
def example (x : Std.U32) : Result Std.U32 := do
  ok x
```

However this is invalid lean code since "example" is reserved. This happens despite Aeneas knowing that "example" is a reserved word. The current «…» escaping is only active for record fields and trait methods. Looks reasonable to do the «…» escaping at the point where name registration does it for those with a hyphen.

### LINKED ISSUE #1250: Bug: Anonymous `const` items (`const _`) generate invalid Lean identifiers
An anonymous `const` item:

```rust
const _: () = ();
```

translates successfully, but the Lean backend emits:

```lean4
@[global_simps, irreducible] def _ : Unit := ()
```

which Lean 4.31.0 rejects with:

```txt
Code/Funs.lean:20:32: error: unexpected token '_'; expected identifier
```

This affects e.g. anonymous compile-time assertions.

This is distinct from #1140 (which was about having a loop inside) and may be related to #1163/#1169 (they are about Rust identifiers, and perhaps the PR could cover `_` as well -- Cc: @oliver-butterley).

**Versions**:
- Aeneas: daa85d7e8940 (i.e. latest as of the time of submission)
- Charon: 0.1.232

oliver-butterley:
In Lean a reserved keyword is illegal only as a bare name. E.g., `def example ...` is not permitted although `def example.foo ...` or `def foo.example ...` or `def «example» ...` are all fine. 

As such, the PR adds a check for such problems and adds «...» when required. It also adds the quotes in some places where not required but they are harmless, specifically we will get `foo.«example»` even though  `foo.example` would be fine (motivated solely by minimising the PR diff). 

The PR adds a lean utility which automatically produces the ocaml file containing all lean keywords to avoid the list going stale in the future. 

The PR also introduces a separation of the two concerns: 
- Lean keywords which are valid identifiers but need to be escaped with  «...»
- Lean identifiers which are introduced in Aeneas and for which collisions need to be avoided

Note: there is no CI check which enforces that the auto-generated ocaml list of lean keywords is up-to-date. This is the same situation as the other Lean utility for standard library models.

fixes #1163
fixes #1250
fixes #1105


tests/src/issue-1250-anonymous-const.rs
```diff
@@ -0,0 +1,10 @@
+//@ [!lean] skip
+
+//! Regression test for issue https://github.com/AeneasVerif/aeneas/issues/1250
+//!
+//! Anonymous constants (`const _`) are permitted in Rust. However, without
+//! escaping `_` isn't a valid Lean identifier. Moreover `_`` is not an
```
ia0: typo
```suggestion
//! escaping `_` isn't a valid Lean identifier. Moreover `_` is not an
```

sonmarcho (approved):

PR #1182, opened 2026-07-02
fix: pass commit sha into nix container when running release workflow

abentkamp:
This PR fixes the issue that all recent `aeneas` releases report `unknown` when running `aeneas -version`. This happens because the nix build happens in a sandbox without access to the git commit. To fix it, we pass the desired version string into nix via environment variables.

In addition, I made a switch to produce different GitHub tags and version strings depending on whether it was a scheduled nightly or a manually triggered build.

I tested the setup here: https://github.com/cryspen/aeneas/actions/runs/27696222469/job/81920104263
It resulted in this build that reports the version correctly: https://github.com/cryspen/aeneas/releases/tag/build-cc3aad2

AI disclaimer: Claude made the initial draft, but I tested and refined it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,14 @@ jobs:
         run: |
           set -eo pipefail
 
-          TAG_NAME="nightly-$(date +'%Y.%m.%d')"
+          SHORT_SHA="$(git rev-parse --short HEAD)"
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            # Manual run
+            TAG_NAME="build-${SHORT_SHA}"
+          else
+            # Scheduled run
+            TAG_NAME="nightly-$(date +'%Y.%m.%d')-${SHORT_SHA}"
+          fi
           echo "tag_name=${TAG_NAME}" >> "$GITHUB_OUTPUT"
 
       - name: Create Release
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
Nadrieril, line 151: Could you add a check here that `aeneas --version` includes the short commit sha?

.github/workflows/release.yml
```diff
@@ -81,10 +88,14 @@ jobs:
...
       - name: Package
+        env:
+          # The release tag reported by `aeneas -version`.
+          # `--impure` is required for nix to read the env var.
+          AENEAS_RELEASE_VERSION: ${{ needs.prepare.outputs.tag_name }}
         run: |
           set -eo pipefail
 
-          nix build -L .#${{ matrix.nix_attr }}
+          nix build -L --impure .#${{ matrix.nix_attr }}
```
Nadrieril, line 98: I prefer to move that comment just above where it it relevant:
```suggestion
          # Pass `--impure` so that nix can read `AENEAS_RELEASE_VERSION`
          nix build -L --impure .#${{ matrix.nix_attr }}
```

.github/workflows/release.yml
Nadrieril: Even better would be to keep the env var we computed above, and here to check that `aeneas --version` returns exactly that text

.github/workflows/release.yml
abentkamp: Done.

.github/workflows/release.yml
abentkamp: Done.

Issue #1130, opened 2026-06-17
Generate `opaque`s instead of `axiom`s


abentkamp:
When a function is excluded via charon's `--opaque`, Aeneas generates a Lean `axiom` for that function. Maybe Aeneas could use Lean's `opaque` command instead, so that the function does not show up in `#print axioms` later?

oliver-butterley:
Any `axiom` in the output of Aeneas is written to `FunsExternal_Template.lean` and these need to be copied across to `FunsExternal.lean`. By design `FunsExternal.lean` isn't overwritten by Aeneas since the typical use case is to replace the axioms with definitions. Equally it is fine to replace them with `opaque`s. So it seems the feature requested here is already possible with minimal effort. Did I miss something?

abentkamp:
That assumes that you use the `--split-files` option. But indeed, `--split-files` is a simple workaround for this issue.

sonmarcho:
I think we should make `--split-file` the default: that's the way the extraction is supposed to be used.

oliver-butterley:
I agree, I would make `--split-files` the default. I would never consider translating with Aeneas without that option and so it didn't even come to mind when I wrote the comment above!

oliver-butterley:
Thinking more, there is a very valid use case that leaves the axioms produced by Aeneas untouched: one has the Aeneas translation with axioms for anything that Charon registered as opaque and for the purposes of verification, spec axioms get added for all these functions. In this case it becomes messy because so many axioms are introduced into the project. 

I see that having the axioms as opaque would make this cleaner. However, is this always going to work? Without providing a definition for the opaque declaration Lean needs to know that the type is inhabited. Can we guarantee this for everything Aeneas produces? 

Another option would be to introduce a new named axiom, e.g., `AeneasAx` which is used as the single axiom for all the opaque functions. This has rather convenient UX in the sense that then print_axioms will tell concisely what axioms were used in the proof of any given statement and also show if that theorem depends on  something translated as opaque.

abentkamp:
I believe we can safely assume that every item produced by aeneas is inhabited. Typically, everything lives in `Result`, which is inhabited. Maybe there are exceptions like `const`s, but the only Rust type that is not inhabited is the never type, which cannot be used for consts, as far as I know.

oliver-butterley:
[...]

PR #1185, opened 2026-07-02
feat: Multi-file extraction.

mpenciak:
This PR changes the implementation of the `-split-files` flag to extract lean files to mirror the structure of the extracted Rust crate more closely. Roughly with one lean file corresponding to a single Rust file. Also adds a `-dump-file-graph` flag to see the structure of the dependencies between the files.

For declarations forming non-trivial dependency cycles between files, they are separated into their own `MergeCrate1Crate2....lean` files (can re-visit the name). The imports are calculated on the Lean side to match the DAG of dependencies.

We also deal with splitting up files based off of sequences of extracted opaque/non-opaque definitions. This will potentially lead to multiple template files being extracted per rust module if the inter-declaration dependencies are particularly complicated.

Note: An issue in charon came up in the course of implementing this feature, where the environment is not set properly for `visit_type_decl`, causing type <-> type dependencies not to show up in the calculation of SCCs. There is a fix in this PR, but it should be upstreamed to a fix in `iter_crate_with_span` in `charon.

AI Disclosure: Claude was significantly involved in the implementation.


documentation/aeneas-overview.md
```diff
@@ -207,8 +207,11 @@ When Aeneas translates a Rust crate, it produces several Lean files:
 | **TypesExternal.lean**             | User-maintained file with type models (never overwritten by Aeneas)                         |
 | **FunsExternal.lean**              | User-maintained file with function models (never overwritten by Aeneas)                     |
 
-> **Tip:** Use the `-split-files` option to have Aeneas generate one file per
-> declaration group as detailed in the above table. This is helpful for large crates.
+> **Tip:** Use the `-split-files` option to have Aeneas generate one Lean module
+> per Rust source file, mirroring the crate's module structure. Source files that
+> form an import cycle are merged into a single module. The older by-kind layout
+> in the table above (one `Types.lean`/`Funs.lean` etc. for the whole crate) is
+> still available as `-split-files-legacy`. Either is helpful for large crates.
```
abentkamp, line 214: Probably `-split-files` or `-split-files-legacy` should actually be the default.

documentation/aeneas-overview.md
```diff
@@ -411,7 +414,10 @@ representation.
 ```
 
 Useful options:
-- `-split-files` — one file per declaration group
+- `-split-files` — one Lean module per Rust source file (mirrors the crate
+  structure); also emits a `Crate.lean` entry point by default (unless `-subdir`)
+- `-split-files-legacy` — one file per declaration kind (the older by-kind split)
+- `-gen-lib-entry` — redundant: `-split-files` already emits the `Crate.lean` entry by default
```
abentkamp: If `-gen-lib-entry` is really redundant, then we should remove it. This documentation is just confusing.

documentation/getting-started.md
```diff
@@ -40,7 +41,10 @@ This creates Lean files under `proofs/MyCrate/Code/`:
 | `TypesExternal_Template.lean`    | Stubs for external types (you complete this)     |
 
 **Key aeneas flags:**
-- `-split-files` — one file per declaration group (recommended for large crates)
+- `-split-files` — one Lean module per Rust source file, mirroring the crate
+  (recommended for large crates; also emits a `MyCrate.lean` entry point by
+  default, unless `-subdir` is given)
+- `-split-files-legacy` — the older by-kind layout used in this tutorial
```
abentkamp, line 47: Since the new file splitting is brand-new, I think showing the old by-kind-splitting in the tutorial for now might be the right choice. But when the option is called `-split-files-legacy`, that feels odd.

I'd suggest to make the old by-kind splitting the default now, and make `-split-files` enable the new fine-grained file-splitting. We might need an option like `-single-file` to keep the old default behavior (Not sure if anyone needs that...).

src/file/FileGraph.ml
```diff
@@ -0,0 +1,226 @@
...
+  | BExternalTypes  (** Referenced external type declarations. *)
+  | BExternalFuns  (** Referenced external functions/globals/traits. *)
+[@@deriving show, ord]
+
+module BucketOrd : Collections.OrderedType with type t = bucket = struct
+  type t = bucket
+
+  let compare = compare_bucket
+  let to_string = show_bucket
+  let pp_t fmt x = Format.pp_print_string fmt (show_bucket x)
```
abentkamp:
```suggestion
  let pp_t = pp_bucket
```
?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,226 @@
...
+  | BExternalFuns -> "<external funs>"
+
+type t = {
+  crate_name : string;
+  item_bucket : bucket AnyDeclIdMap.t;
+      (** Every declaration's bucket, computed once. Items whose metadata can't
+          be found (e.g. ids a prepass removed from the maps but not from
+          [crate.declarations]) are absent. *)
+  members : item_id list BucketMap.t;
+  edges : BucketSet.t BucketMap.t;
```
abentkamp: A comment saying which direction the edges are pointing would be helpful here. That is, is this the set of dependencies or the set of dependent buckets?

src/file/FileMapping.ml
```diff
@@ -0,0 +1,77 @@
...
+  (* Split into path components, dropping empty and "." segments. *)
+  let parts =
+    String.split_on_char '/' path |> List.filter (fun s -> s <> "" && s <> ".")
+  in
+  (* Drop a leading "src" component. *)
+  let parts =
+    match parts with
+    | "src" :: rest -> rest
+    | _ -> parts
+  in
```
abentkamp: I ran into an issue with this when trying to extract code in a Rust workspace. Since the paths provided by charon are relative the the workspace root, the `src` folder then appears in the middle of the path.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,226 @@
...
+(** The bucket an item belongs to, or [None] if we can't find its metadata.
+
+    Bucketing is gated on [is_local]: a local item goes to a [BFile] keyed on
+    its source path, a non-local item to an external bucket by kind. *)
+let bucket_of_item (crate : crate) (id : item_id) : bucket option =
+  match crate_get_item_meta crate id with
+  | None -> None
+  | Some meta ->
+      if meta.is_local then
+        Some (BFile (Meta.path_of_file_name meta.span.data.file.name))
```
abentkamp: Using the `span` as the source of truth for assigning the bucket doesn't work, as I found out running this on a larger crate. For example, when the item is defined using a macro from an external crate, the span is pointing to the external crate, but the item is still local. This completely breaks the extraction.

oliver-butterley (commented):
Here are a few initial comments / questions:

- Did I understand correctly that the `-gen-lib-entry` flag now does nothing? Would it be better to remove this?
- I see the new tests don't use non-builtin external dependencies, would it be possible to add a test which also tests this functionality? (Particularly since it seems the case that is more precarious.) 
- Would it be possible for `noncomputable` to be set per file rather than a single setting for the entire crate? Not essential but would be convenient sometimes if it were easy.
- Does the code of `module_components_of_file` make too strong assumptions on what the received paths will look like? Isn't is possible that a path could contain `..` and this isn't handled?
- Could the comment for `iter_crate_with_span` include also a link to the Charon issue related to this to better track future patching?

src/file/FilePlan.ml
```diff
@@ -0,0 +1,373 @@
...
+            match runs_rev with
+            | (t', gs) :: rest when t' = t -> ((t, g :: gs) :: rest, pending_rev)
+            | _ -> ((t, g :: pending_rev) :: runs_rev, [])))
+      ([], []) groups
+  in
+  match (runs_rev, pending_rev) with
+  | [], [] -> []
+  | [], pending_rev -> [ (false, List.rev pending_rev) ]
+  | runs_rev, _ ->
+      List.rev_map (fun (t, gs_rev) -> (t, List.rev gs_rev)) runs_rev
```
abentkamp: It's hard to see what's happening to `pending_rev`. This might help:
```suggestion
  | runs_rev, pending_rev ->
      assert (pending_rev = []);
      List.rev_map (fun (t, gs_rev) -> (t, List.rev gs_rev)) runs_rev
```

src/file/FilePlan.ml
```diff
@@ -0,0 +1,373 @@
...
+            [ suffix ]
+          else
+            match buckets with
+            | [ BFile p ] -> FileMapping.module_components_of_file p
+            | _ ->
+                let paths =
+                  List.filter_map
+                    (function
+                      | BFile p -> Some p
+                      | BExternalTypes | BExternalFuns -> None)
```
abentkamp: If `BExternalTypes` or `BExternalFuns` is part of a multi-bucket SCC, something must have gone seriously wrong, right? Maybe we could throw an error here?

src/Translate.ml
```diff
@@ -1436,6 +1440,224 @@ let extract_file (config : gen_config) (ctx : gen_ctx) (fi : extract_file_info)
...
+                  filename = layer.filename;
+                  namespace;
+                  in_namespace = not m.is_external;
+                  (* External modules don't open the crate namespace: their
+                     declarations have absolute (external) names, and they
+                     cannot import a local module to obtain the namespace
+                     without creating an import cycle (every local module
+                     imports the external modules).
+
+                     TODO: We will have to revisit this in a bit
```
abentkamp: What is this TODO about?

src/Translate.ml
```diff
@@ -2123,18 +2376,11 @@ let extract_translated_crate (filename : string) (dest_dir : string)
...
-        (* Write *)
-        Printf.fprintf out "import %s.Funs\n" crate_name;
-        (* Flush and close the file, log *)
-        close_out out;
-        log#linfo (lazy ("Generated: " ^ filename)));
+      (* Note: the [-split-files] entry point ([Crate.lean]) is generated inside
+         [extract_by_file]. The legacy split ([-split-files-legacy]) has no entry
+         point: it cannot produce a buildable one without [-subdir] (which is
+         incompatible with [-gen-lib-entry]), so that combination is rejected at
+         the CLI. *)
```
abentkamp: I don't fully understand this comment, but I don't think this is the right place for it.

abentkamp (commented):
Thanks for implementing this! At Cryspen, we wanted to have this feature for a while now but never got around to implement it.

I am not sure if the Part/Axiom splitting is the right approach when the user marks some items as opaque. What I would expect when marking an item as opaque is that this translates into a Lean item using Lean's `opaque`. Then the user can provide an axiomatized spec for that item to model what it is supposed to do. This is related to what I wrote here: https://github.com/AeneasVerif/aeneas/issues/1130. With this approach, splitting files into Part/Axiom is completely unnecessary, and it's clear what is extracted and what is specified manually.

I've run the PR on a larger crate and I found two issues:
- `FunsExternal` and `TypesExternal` gets split into Axioms/Parts, too. Is that intentional? I think it's because some external items can have a translatable body.
- When using external macros, the file splitting breaks (see one of the comments below).

I also have a couple of smaller comments below.

mpenciak:
> Here are a few initial comments / questions:
> 
> * Did I understand correctly that the `-gen-lib-entry` flag now does nothing? Would it be better to remove this?

I think `-gen-lib-entry` still makes a `Crates.lean` file in the `-split-files-legacy` mode, but I think it's reasonable to make it the default behavior for `-split-files-legacy` as well (as it is currently the default behavior for `-split-files`).

> * I see the new tests don't use non-builtin external dependencies, would it be possible to add a test which also tests this functionality? (Particularly since it seems the case that is more precarious.)

That's a good catch, I'll try to find a way to write a small test that works 

> * Would it be possible for `noncomputable` to be set per file rather than a single setting for the entire crate? Not essential but would be convenient sometimes if it were easy.

I don't think this should be terribly difficult to implement

> * Does the code of `module_components_of_file` make too strong assumptions on what the received paths will look like? Isn't is possible that a path could contain `..` and this isn't handled?

You're right about that as well, I'll see if there's a better way to implement the file path splitting.

> * Could the comment for `iter_crate_with_span` include also a link to the Charon issue related to this to better track future patching?

I haven't made an issue in `charon` yet for this, but I'll do that now and link it 

So tl;dr on the TODOs from your comment:

- [ ] Delete `-gen-lib-netry` and make it the default behavior for both the file splitting modes
- [ ] Add a test exercising importing external crates with the file splitting logic
- [ ] Change the `noncomputable` behavior to be per-file (This will probably be the biggest change, so if it looks harder than I expect I might leave it off to a future PR)
- [ ] Fix up file splitting in `module_components_of_file`
- [ ] Make a `charon` issue for the type <-> type dependency graphs and link it in `Deps.ml`

mpenciak:
> Thanks for implementing this! At Cryspen, we wanted to have this feature for a while now but never got around to implement it.
> 
> I am not sure if the Part/Axiom splitting is the right approach when the user marks some items as opaque. What I would expect when marking an item as opaque is that this translates into a Lean item using Lean's `opaque`. Then the user can provide an axiomatized spec for that item to model what it is supposed to do. This is related to what I wrote here: #1130. With this approach, splitting files into Part/Axiom is completely unnecessary, and it's clear what is extracted and what is specified manually.
> 

So my reasoning for this is the following: 

I think the usual workflow with Aeneas is you extract (or better: You have re-extract as part of your CI), and any file that's generated by Aeneas (and labeled as such) should be checked in CI to have not changed from the original extraction. 

For `_template` files in the legacy mode/`Axiom` files in the new mode the user will want to fill in the gaps with their own models, so you don't want to re-check those in CI. In the legacy mode, all the axioms/opaques were put into the `{Funs,Types}External_template.lean` and we called it a day, but when splitting by file we'd need a separate `_template` file for each file with an opaque.

And then when you have non-opaques and opaques depending on each other... That's when this splitting becomes an issue and in order for this separation of `_template` files to work we need to have this multi-step extraction.

The example in the test is really contrived, and I expect this kind of splitting will almost never happen. Usually the opaques will be concentrated in external dependencies, which are imported everywhere.

> I've run the PR on a larger crate and I found two issues:
> 
> * `FunsExternal` and `TypesExternal` gets split into Axioms/Parts, too. Is that intentional? I think it's because some external items can have a translatable body.

Oh yeah, that's definitely broken. I'll go in and fix that as well!

> * When using external macros, the file splitting breaks (see one of the comments below).
> 

I'll look into that! Thanks for finding that breakage.

> I also have a couple of smaller comments below.

I'll start working on addressing those as well! Thanks for your review!

abentkamp:
> So my reasoning for this is the following:
>
> I think the usual workflow with Aeneas is you extract (or better: You have re-extract as part of your CI), and any file that's generated by Aeneas (and labeled as such) should be checked in CI to have not changed from the original extraction.
>
> For _template files in the legacy mode/Axiom files in the new mode the user will want to fill in the gaps with their own models, so you don't want to re-check those in CI. In the legacy mode, all the axioms/opaques were put into the {Funs,Types}External_template.lean and we called it a day, but when splitting by file we'd need a separate _template file for each file with an opaque.
>
> And then when you have non-opaques and opaques depending on each other... That's when this splitting becomes an issue and in order for this separation of _template files to work we need to have this multi-step extraction.
>
> The example in the test is really contrived, and I expect this kind of splitting will almost never happen. Usually the opaques will be concentrated in external dependencies, which are imported everywhere.

Okay, you are right, this is consistent with how Aeneas works. I am not convinced that this is the right way, but that's a question we should discuss independently of this PR.

We should probably rethink the name `Axioms` since only the template contains axioms, but they are supposed to be replaced by a definition, right? So maybe they should be called `HandWrittenPart` or `MockPart`? And maybe `Part` should be called `ExtractedPart`?

I'd be happy with the name `Merge`, but maybe `Bundle` would be more fitting? That's what we call them in Hax.

documentation/aeneas-overview.md
sonmarcho: I agree: `-split-files` should be the default. Then we can have an option `-no-split-files` that will mostly be used in the Aeneas tests (no need to mention it here). I think we also don't need to mention the `-split-files-legacy` in the document: it should never be used for new projects.

documentation/getting-started.md
sonmarcho: Using `-split-files-legacy` shouldn't make much difference on small crates, so I think we can use the default `-split-files` in the tutorial should work?

README.md
```diff
@@ -152,16 +152,21 @@ and tactics specialized for monadic programs (see
...
 For instance, it can lead to the following structure in Lean:
 ```
-... // Files containing type definitions, etc.
-FunsExternal_Template.lean // automatically generated template file
-FunsExternal.lean // hand-written file maintained by the user (not overwritten during the translation)
-Funs.lean // automatically generated file which imports FunsExternal.lean
+Crate.lean // automatically generated library entry point
+Crate/Foo.lean, Crate/Bar.lean, ... // one automatically generated module per Rust source file
+Crate/FunsExternal_Template.lean // automatically generated template file
+Crate/FunsExternal.lean // hand-written file maintained by the user (not overwritten during the translation)
```
sonmarcho: I think we want to explain the case where we generate more than one file (generating a single `FunsExternal.lean` is too close to the legacy behavior).

README.md
```diff
@@ -152,16 +152,21 @@ and tactics specialized for monadic programs (see
 
 When translating a crate which requires external definitions (i.e., definitions
 coming from external dependencies such as the Rust standard library), we advise using the
-`-split-files` option. With `-split-files`, Aeneas will generate template files listing
-the missing definitions for which the user will have to provide hand-written models.
+`-split-files` option. With `-split-files`, Aeneas generates one Lean module per Rust
+source file (mirroring the crate structure), plus template files listing the missing
+external definitions for which the user has to provide hand-written models.
```
sonmarcho, line 157: We should be more explicit about the fact that "external" refers to "external dependencies"

tests/src/split_files/macrolib/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[macro_export]
+macro_rules! define_counter {
+    ($name:ident) => {
+        pub struct $name {
+            pub value: u32,
+        }
+
+        impl $name {
+            pub fn bump(&mut self) {
```
sonmarcho, line 13: nit: what about `incr` rather than `bump`?

sonmarcho (commented):
I'm doing the review in several steps.
First, a few comments about the design choices:
- I probably suggested `Merge`, but `Bundle` may indeed be a better name
- I would rather put `Bundle` at the end of the name `BundlePingPong` -> `PingPongBundle` so that we avoid grouping all the bundles together when listing the files alphabetically - here the bundle would be put after `Ping` (of course we might want to put it after `Pong`, so there is an arbitrary choice here). What do you think (it's merely a suggestion: I don't have a strong opinion)?
- `Axioms` may indeed be too strong of a name (though from the point of view of verification a model is essentially an axiom). What about `Opaque`?

I also put some remarks as comments

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+open LlbcAst
+open LlbcAstUtils
+open Types
+open Meta
+
+(** Where a declaration/reference lands in the projected file graph.
+
+    The external buckets are diagnostic buckets for non-local referenced
+    declarations. They do not imply that the current extraction backend will
+    emit a corresponding external template file. *)
```
sonmarcho, line 14: I don't fully understand this comment. Does it mean that we might skip the "external" buckets if all of their declarations have models in the standard Aeneas library?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+
+(** The source path a local item's Rust *module* path implies, under the usual
+    [crate::a::b -> src/a/b.rs] convention. Used only as a fallback when the
+    item's span is unusable (see [bucket_of_item]). *)
+let source_path_of_name (name : Types.name) : string =
+  (* A trailing [PeTarget] (multi-target mode) would otherwise end the run of
+     identifiers early and make us mistake the item's own name for a module. *)
+  let name =
+    match strip_target_suffix name with
+    | _crate :: rest -> rest
```
sonmarcho, line 45: We have to be careful: `_crate` does not always refer to the local crate

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+    | [] -> []
+  in
+  (* The leading run of plain identifiers. *)
```
sonmarcho, line 48: AI generated comment (I'm not going to list them all)

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+
+module BucketSet = Collections.MakeSet (BucketOrd)
+module BucketMap = Collections.MakeMap (BucketOrd)
+
+(** The source path a local item's Rust *module* path implies, under the usual
+    [crate::a::b -> src/a/b.rs] convention. Used only as a fallback when the
+    item's span is unusable (see [bucket_of_item]). *)
+let source_path_of_name (name : Types.name) : string =
+  (* A trailing [PeTarget] (multi-target mode) would otherwise end the run of
+     identifiers early and make us mistake the item's own name for a module. *)
```
sonmarcho, line 42: I don't get this

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  let pp_t = pp_bucket
+  let show_t = show_bucket
+end
+
+module BucketSet = Collections.MakeSet (BucketOrd)
+module BucketMap = Collections.MakeMap (BucketOrd)
+
+(** The source path a local item's Rust *module* path implies, under the usual
+    [crate::a::b -> src/a/b.rs] convention. Used only as a fallback when the
+    item's span is unusable (see [bucket_of_item]). *)
```
sonmarcho, line 39: Could you make a pass over the comments? They seem AI generated and are sometimes a bit cryptic.

Also I don't get this: I think an example would be useful.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  in
+  (* The leading run of plain identifiers. *)
+  let rec idents (n : Types.path_elem list) : string list =
+    match n with
+    | PeIdent (s, _) :: rest -> s :: idents rest
+    | _ -> []
+  in
+  let prefix = idents name in
+  (* If no impl element cut the run short, the last identifier is the item's own
+     name rather than a module. *)
```
sonmarcho, line 56: AI comment, etc.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  | [] -> "src/lib.rs"
+  | _ -> "src/" ^ String.concat "/" modules ^ ".rs"
+
+(** The bucket an item belongs to, or [None] if we can't find its metadata.
+
+    Bucketing is gated on [is_local]: a local item goes to a [BFile] keyed on
+    its source path, a non-local item to an external bucket by kind.
+
+    [on_foreign_span] is called for a local item whose span points into another
+    crate. *)
```
sonmarcho, line 74: AI comment, and also: it seems obsolete (I don't see `is_local`, and `on_source` is not explained).
Also it doesn't explain what `on_foreign_span` should do.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+           the same key for items of the same file to share a bucket. *)
+        if file.crate_name = crate.name then (
+          let b = BFile (FileMapping.normalize_source_path path) in
+          on_source b path;
+          Some b)
+        else (
+          (* An item generated by a `macro_rules!` macro from another crate is
+             local, but its span points at the macro *definition* — a file of
+             the defining crate, which may not even appear in this crate's file
+             table.
```
sonmarcho, line 100: I know there are restrictions with regards to the use of macros so maybe this can't happen, but what about defining a macro in a crate and using it in the same crate but in a different module? Could this lead to an item being placed in the wrong bucket?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+             taking the span at face value files it under a foreign path,
+             emitting a module that mirrors the other crate's tree.
+
+             So fall back to the Rust module path, which is always local and,
+             under the usual module-to-file convention, names the file the macro
+             was invoked from. That convention is not universal ([#[path]],
+             inline [mod] blocks, and [mod.rs] all break it), so the item may
+             land in a module of its own rather than merging with its
+             file-mates — still a local, sensible tree, unlike the
+             alternative. *)
```
sonmarcho, line 117: AI slop again

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+          let b = BFile (FileMapping.normalize_source_path reconstructed) in
+          on_source b reconstructed;
+          Some b)
+
+type t = {
+  crate_name : string;
+  item_bucket : bucket AnyDeclIdMap.t;
+      (** Every declaration's bucket, computed once. Items whose metadata can't
+          be found (e.g. ids a prepass removed from the maps but not from
+          [crate.declarations]) are absent. *)
```
sonmarcho, line 129: "computed once": what does this mean?
Also what do we do for items with no metadata? Are they simply ignored?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+      (** [edges] maps a bucket to the buckets it uses. Self-edges are omitted.
+      *)
+  sccs : bucket SCC.sccs;
+}
+
+let compute (crate : crate) : t =
+  (* Enumerate the declarations through [crate.declarations] — the same source
+     of truth the emitter partitions — and map each item to its bucket once. *)
+  (* [declaration_group_to_list] comes from charon's [GAstUtils] via
+     [LlbcAstUtils]. *)
```
sonmarcho, line 145: Remove this

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  members : item_id list BucketMap.t;
+  edges : BucketSet.t BucketMap.t;
+      (** [edges] maps a bucket to the buckets it uses. Self-edges are omitted.
+      *)
+  sccs : bucket SCC.sccs;
+}
+
+let compute (crate : crate) : t =
+  (* Enumerate the declarations through [crate.declarations] — the same source
+     of truth the emitter partitions — and map each item to its bucket once. *)
```
sonmarcho, line 143: "the same source of truth the emitter partitions": AI slop

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  (* [declaration_group_to_list] comes from charon's [GAstUtils] via
+     [LlbcAstUtils]. *)
+  let items = List.concat_map declaration_group_to_list crate.declarations in
+  (* Local items whose span pointed into another crate, counted per offending
+     file so the workaround in [bucket_of_item] reports once per file instead of
+     once per item. *)
+  let foreign_spans : (string, int) Hashtbl.t = Hashtbl.create 4 in
```
sonmarcho, line 150: Why do we need this? Also: what's the `int` for? Can we simply use a set?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  let on_foreign_span (path : string) : unit =
+    Hashtbl.replace foreign_spans path
+      (1 + Option.value (Hashtbl.find_opt foreign_spans path) ~default:0)
+  in
```
sonmarcho, line 154: What does this do?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  (* The path charon reported for each file bucket, kept for diagnostics. All
+     items of a bucket agree on it, bar the [mod.rs]-style imprecision of the
+     macro fallback, where first-seen is as good as any. *)
+  let bucket_source : string BucketMap.t ref = ref BucketMap.empty in
+  let on_source (b : bucket) (path : string) : unit =
+    if not (BucketMap.mem b !bucket_source) then
+      bucket_source := BucketMap.add b path !bucket_source
+  in
```
sonmarcho, line 162: Similar questions

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+      (fun acc id ->
+        match bucket_of_item crate ~on_foreign_span ~on_source id with
+        | None ->
+            (* No metadata: a prepass removed the id from the decl maps but not
+               from [crate.declarations]. The item can't be bucketed, so it is
+               dropped from the split output ([FilePlan.place_by_file] also
+               warns when this drops a whole declaration group). *)
+            [%warn_opt_span] None
+              ("Multi-file extraction: no metadata found for " ^ show_item_id id
+             ^ "; it will not appear in the output");
```
sonmarcho, line 174: So the answer to one of the above questions is here: we print a warning and don't extract the item. It's not necessarily a bad choice (it's very unlikely that Charon fails to retrieve the span of an item), but it should be properly documented.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+  Hashtbl.iter
+    (fun path n ->
+      [%warn_opt_span] None
+        (Printf.sprintf
+           "Multi-file extraction: %d local declaration(s) have a span \
+            pointing into %s, which is not a file of this crate. Most likely \
+            they come from a macro defined in another crate. Placing them by \
+            their Rust module path instead of by that file; the resulting \
+            module may not merge with the rest of the file the macro was \
+            invoked from."
```
sonmarcho, line 188: I don't think we need to warn: I expect this to be common.
But I want to investigate: I think the spans give us both the span at which the macro is defined *but also* the span at which it is used - that is we can retrieve the span of the definition. @Nadrieril should know more.

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+          canonical bucket key has stripped to the crate's source root. Only for
+          diagnostics: it is what the user can actually find on disk. *)
+  members : item_id list BucketMap.t;
+  edges : BucketSet.t BucketMap.t;
+      (** [edges] maps a bucket to the buckets it uses. Self-edges are omitted.
+      *)
+  sccs : bucket SCC.sccs;
+}
+
+let compute (crate : crate) : t =
```
sonmarcho, line 141: This function probably deserves a (short) docstring

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+let display_bucket (g : t) (b : bucket) : string =
+  match b with
+  | BFile p -> Option.value (BucketMap.find_opt b g.bucket_source) ~default:p
+  | BExternalTypes -> "<external types>"
+  | BExternalFuns -> "<external funs>"
+
+(** Render the report.
+
+    [get_name] turns an item id into a human-readable Rust name. *)
+let render (graph : t) ~(get_name : item_id -> string) : string =
```
sonmarcho, line 268: What about `graph_to_string`?

src/file/FileGraph.ml
```diff
@@ -0,0 +1,339 @@
...
+    bucket_source = !bucket_source;
+    members = !members;
+    edges = !edges;
+    sccs;
+  }
+
+(** How to name a bucket to the user: for a file, the path charon reported,
+    which is what they can find on disk. Falls back to the canonical key for a
+    bucket the graph doesn't know (it always does, in practice). *)
+let display_bucket (g : t) (b : bucket) : string =
```
sonmarcho, line 259: What about `bucket_to_string`? That's the convention used in the project (`display` evokes something effectful)

src/file/FileMapping.ml
```diff
@@ -0,0 +1,115 @@
...
+
+    Conventions:
+    - ["src/foo.rs"] -> [["Foo"]]
+    - ["src/baz/bang.rs"] -> [["Baz"; "Bang"]]
+    - ["src/geometry/mod.rs"] -> [["Geometry"; "Mod"]]
+    - ["src/geometry.rs"] -> [["Geometry"]]
+    - ["src/lib.rs"] -> [["Lib"]]
+    - ["src/main.rs"] -> [["Main"]]
+    - ["src/cycle_x.rs"] -> [["CycleX"]] (snake_case -> CamelCase)
+    - ["crates/mycrate/src/foo.rs"] -> [["Foo"]] (workspace member) *)
```
sonmarcho, line 20: Doesn't the workspace member case lead to collisions?

src/file/FileMapping.ml
```diff
@@ -0,0 +1,115 @@
...
+     reduces to nothing means a broken upstream invariant. *)
+  if parts = [] then
+    [%craise_opt_span] None
+      ("Cannot map the source file path to a Lean module: " ^ path);
+  List.map StringUtils.to_camel_case parts
+
+(** The module-path components for a merged (multi-file) SCC.
+
+    The name is derived from the member files: each path is camel-cased like a
+    single-file module and the results are concatenated in alphabetical order *)
```
sonmarcho, line 96: About the naming of bundled multi-file SCCs: maybe we want to use a delimiter like "_" between the different module names

src/utils/Deps.ml
```diff
@@ -68,7 +68,16 @@ let compute_graph_of_uses (m : crate) : graph_of_uses =
...
-      inherit [_] iter_crate_with_span
+      inherit [_] iter_crate_with_span as super
+
+      (* Workaround: upstream `iter_crate_with_span` does not track the current
+         item environment for type declarations, making type-to-type uses invisible.
+         Override here until fixed upstream:
+         https://github.com/AeneasVerif/charon/issues/1346 *)
+      method! visit_type_decl (_ : (item_id * span) option) (decl : type_decl) =
+        let decl_span_info = Some (IdType decl.def_id, decl.item_meta.span) in
+        super#visit_type_decl decl_span_info decl
```
sonmarcho, line 79: I see the issue has been closed

src/Config.ml
```diff
@@ -216,13 +216,21 @@ let always_deconstruct_adts_with_matches = ref false
...
 let split_files = ref false
 
-(** Only for Lean: generate the library entry point, if the crate is split
-    between different files. The entry point is simply a file with the name of
-    the crate and which includes all the other files. *)
+(** Only for Lean, and only with [-split-files-legacy]: generate a library entry
+    point — a file named after the crate that imports the split modules. The
+    per-file [-split-files] mode emits its entry point by default, so this flag
+    does not apply there. *)
 let generate_lib_entry_point = ref false
```
sonmarcho, line 234: As discussed, should be true by default.
I also think we need to update the docstring.

src/Config.ml
```diff
@@ -216,13 +216,21 @@ let always_deconstruct_adts_with_matches = ref false
...
+(** Legacy split mode: split the generated definitions between different files
+    by kind (types, clauses and functions), instead of grouping them in one
+    file. Superseded by [split_files]; kept as [-split-files-legacy]. *)
+let split_files_legacy = ref false
+
+(** Split the generated definitions into one module per Rust source file,
+    mirroring the crate structure. Source files that form an import cycle are
+    merged into a single module. Mutually exclusive with [split_files_legacy].
+*)
 let split_files = ref false
```
sonmarcho, line 228: As discussed, should be true by default

src/Config.ml
```diff
@@ -216,13 +216,21 @@ let always_deconstruct_adts_with_matches = ref false
 (** Controls whether we use fuel to control termination. *)
 let use_fuel = ref false
 
-(** Controls whether we split the generated definitions between different files
-    for the types, clauses and functions, or if we group them in one file. *)
+(** Legacy split mode: split the generated definitions between different files
+    by kind (types, clauses and functions), instead of grouping them in one
+    file. Superseded by [split_files]; kept as [-split-files-legacy]. *)
```
sonmarcho, line 221: I think it should be "intead of grouping them by module" (or "source file")?

src/EmitJson.ml
```diff
@@ -272,12 +269,20 @@ let trait_impl_entry_of_trait_impl (ctx : ExtractBase.extraction_ctx)
 
 let begin_file_if_enabled ~(filename : string) ~(namespace : string) : unit =
   if !Config.emit_json then begin
-    (* Record the Lean file relative to dest_dir. *)
-    let basename = Filename.basename filename in
+    (* Record the Lean file relative to [dest_dir]. With [-split-files] the
+       relative path can have several directory components (e.g.
+       [Crate/Alpha/Part1.lean]), so strip the [dest_dir] prefix instead of
+       taking the basename. *)
```
sonmarcho, line 275: I don't fully get this

src/Main.ml
```diff
@@ -465,6 +473,16 @@ let () =
       "The -max-recdepth option is valid only for the Lean backend";
   if !emit_json && not (backend () = Lean) then
     fail_with_error "The -emit-json option is valid only for the Lean backend";
+  check_arg_not !split_files_legacy "-split-files-legacy" !split_files
+    "-split-files";
+  if !split_files && not (backend () = Lean) then
+    fail_with_error
+      "The -split-files option is valid only for the Lean backend. For the \
+       by-kind split (Types/Funs/...), use -split-files-legacy";
```
sonmarcho, line 481: I think we should do `-split-files` by default for Lean and `-split-files-legacy` by default for the other backends

src/llbc/LlbcAstUtils.ml
```diff
@@ -84,6 +84,86 @@ let crate_has_opaque_non_builtin_decls (k : crate) (filter_builtin : bool)
...
+
+    Same predicates as {!crate_get_opaque_non_builtin_decls} (which only covers
+    types and funs); a global delegates to its initializer's body. An item whose
+    declaration can't be found is treated as transparent. *)
+let item_is_opaque (k : crate) (id : item_id) : bool =
+  let fun_is_opaque (fid : FunDeclId.id) : bool =
+    match FunDeclId.Map.find_opt fid k.fun_decls with
+    | None -> false
+    | Some d -> not (body_is_known d.body)
+  in
```
sonmarcho, line 151: nit: why is this one defined separately (and not the others)?

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
+(** Placement plan for [-split-files]: which Lean modules/files each SCC of the
+    file-dependency graph becomes, including the per-module opacity layering .
+    Used by the emitter ({!Translate.extract_by_file}) and the
+    [-dump-file-graph] flag. *)
+
+open FileGraph
+
+(** One emitted declaration file of a placed module: a regenerated [PartK] file,
+    or a user-fillable [AxiomsK_Template] file holding an opaque layer. *)
+type placed_layer = {
```
sonmarcho, line 10: I think we can find a better name (I don't fully get "placed" - but I'm fine with "layer")?
What about simply something like `decl_file`/`extract_file`?
Or maybe it's just about explaining what "placed" refers to.

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+          same module, or in any module it transitively imports. Always [false]
+          for a template, which holds axioms and no definitions. *)
+}
+
+(** One generated Lean module in [-split-files] mode: a source file, a merged
+    SCC, or an external bucket. A module whose declarations are uniform in
+    opacity is a single file at its plain name; a module whose declarations
+    alternate splits into a chain of layers plus an imports-only aggregator at
+    the plain name. *)
+type placed_module = {
```
sonmarcho, line 31: I don't fully get why we need modules and files and I actually expected that a file might contain several modules (not the other way around). I think it would be good to document that.

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+  | None -> Filename.concat full_dest_dir crate_name
+
+(** The opacity "color" of a declaration group.
+
+    All-builtin groups are colorless: they emit nothing, so they attach to
+    whichever layer is current instead of cutting one. A group mixing opaque and
+    transparent members must stay whole, so it degrades to a transparent
+    (regenerated) layer — its opaque members are emitted as inline axioms — with
+    a warning. *)
+type group_color = Colorless | ColorOpaque | ColorTransparent
```
sonmarcho, line 71: Could we use a different name than "color"? Transparency would probably be more appropriate for instance. Or visibility perhaps.

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+            Option.map
+              (fun (m : Types.item_meta) -> m.span)
+              (LlbcAstUtils.crate_get_item_meta crate first)
+          in
+          [%warn_opt_span] span
+            ("Mutually-recursive declaration group mixes opaque and \
+              transparent members: the opaque members are emitted as inline \
+              axioms in a regenerated file instead of a user-fillable \
+              template. Opaque members: "
+            ^ String.concat ", " (List.map Types.show_item_id opaque)));
```
sonmarcho, line 99: This is useful, but I think it may duplicate a similar error message elsewhere: it would be good to make sure we only have one.

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+      ([], []) groups
+  in
+  (* The first colored group flushes [pending_rev] into the run it opens, so a
+     non-empty [pending_rev] survives only when no group was colored at all. *)
+  match (runs_rev, pending_rev) with
+  | [], [] -> []
+  | [], pending_rev -> [ (false, List.rev pending_rev) ]
+  | runs_rev, pending_rev ->
+      [%sanity_check_opt_span] None (pending_rev = []);
+      List.rev_map (fun (t, gs_rev) -> (t, List.rev gs_rev)) runs_rev
```
sonmarcho, line 132: This needs explanations as to how the algorithm works (I don't understand it :P )

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+               (* The group-level counterpart of the missing-metadata warning
+                  in [FileGraph.compute]. *)
+               let repr =
+                 match LlbcAstUtils.declaration_group_to_list g with
+                 | [] -> "<empty declaration group>"
+                 | repr :: _ -> Types.show_item_id repr
+               in
+               [%warn_opt_span] None
+                 ("Multi-file extraction: dropping a declaration group whose \
+                   representative has no source-file bucket (representative: "
```
sonmarcho, line 179: This message may not be very clear to the user

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+     share a source file / SCC, so any member is a valid representative; a
+     group whose representative has no bucket (no metadata) maps to no SCC and
+     is dropped, with a warning. *)
+  let bucket_to_scc =
+    SCC.SccId.Map.fold
+      (fun scc_id buckets acc ->
+        List.fold_left (fun acc b -> BucketMap.add b scc_id acc) acc buckets)
+      fg.sccs.sccs BucketMap.empty
+  in
+  let group_scc (g : LlbcAst.declaration_group) : SCC.SccId.id option =
```
sonmarcho, line 157: I think we need a comment (this lookups the id of the SCC the declaration group belongs to right?)

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+    match LlbcAstUtils.declaration_group_to_list g with
+    | [] -> None
+    | repr :: _ ->
+        Option.bind (LlbcAstUtils.AnyDeclIdMap.find_opt repr fg.item_bucket)
+          (fun b -> BucketMap.find_opt b bucket_to_scc)
+  in
+  let groups_by_scc : LlbcAst.declaration_group list SCC.SccId.Map.t =
```
sonmarcho, line 164: I think we also need a comment: this computes the map from SCC id to the list of declaration groups it contains I think?

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+               (SCC.SccId.Map.find_opt scc_id fg.sccs.scc_deps)
+               ~default:SCC.SccId.Set.empty)
+        in
+        SCC.SccId.Map.add scc_id
+          (from_deps || List.exists group_has_axioms groups)
+          acc)
+      SCC.SccId.Map.empty scc_list
+  in
+
+  (* Does this SCC's *first* layer already see an axiom through its imports? *)
```
sonmarcho, line 223: I think it would be good to explain why we need this: I see it is related to `runs_noncomputable`, but I don't fully get the logic behind the computation of `noncomputable`

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+        let groups =
+          Option.value (SCC.SccId.Map.find_opt scc_id groups_by_scc) ~default:[]
+        in
+        let runs =
+          if is_dropped then []
+          else if is_external then external_run groups
+          else cut_layers color groups
+        in
+        let runs_nc = runs_noncomputable ~scc_id runs in
+        let aggregator, layers =
```
sonmarcho, line 337: I think it would be good to explain the logic of what's happening here

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+  let file_of_components ~(is_template : bool) (components : string list) :
+      string =
+    module_root_dir ^ "/"
+    ^ String.concat "/" components
+    ^ (if is_template then "_Template" else "")
+    ^ ext
+  in
+  let placed =
+    List.map
+      (fun (scc_id, buckets) ->
```
sonmarcho, line 270: The function in the `List.map` is big and at a high level of indentation: it's probably worth introducing a (well-documented) auxiliary function.

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+            | Some ((prev : placed_module), _) ->
+                [%craise_opt_span] None
+                  (Printf.sprintf
+                     "Module-name collision in -split-files: %s and %s both \
+                      map to the Lean module [%s] (file %s). Rename one of the \
+                      Rust sources so their camel-cased module names differ."
+                     (describe prev) (describe m) import_name filename)
+            | None -> Collections.StringMap.add import_name (m, filename) seen)
+          seen entries)
+      Collections.StringMap.empty placed
```
sonmarcho, line 429: What about just using an index to differentiate the files (I'm leaning more and more towards "let's extract something no matter what") while eventually printing a warning saying that there is a name collision and we had to arbitrarily disambiguate the names?

src/file/FilePlan.ml
```diff
@@ -0,0 +1,473 @@
...
+  in
+  placed
+
+(** Render the generated-files section of the [-dump-file-graph] report: exactly
+    the files the [-split-files] emitter will write (dropped builtin-only
+    external modules excluded), with the opacity layering. *)
+let render_placement (placed : placed_module list) : string =
```
sonmarcho, line 436: I suggest `..._to_string`

src/Translate.ml
```diff
@@ -1436,6 +1440,213 @@ let extract_file (config : gen_config) (ctx : gen_ctx) (fi : extract_file_info)
...
+      (fun acc (m : FilePlan.placed_module) -> SCC.SccId.Map.add m.scc_id m acc)
+      SCC.SccId.Map.empty placed
+  in
+
+  let base_config =
+    {
+      (full_extraction_config ()) with
+      (* The CLI rejects [-split-files -decreases-clauses]. Hardcode [false]
+         instead of reading the config refs, so that lifting the CLI rejection
+         cannot silently half-enable an unimplemented feature here. *)
```
sonmarcho, line 1483: I wrote above I would not flag all AI comments, but I can't resist flagging this one ^^

src/Translate.ml
```diff
@@ -1436,6 +1440,213 @@ let extract_file (config : gen_config) (ctx : gen_ctx) (fi : extract_file_info)
...
+    let out = open_out filename in
+    Printf.fprintf out "%s\n" generated_header;
+    Printf.fprintf out "-- [%s]%s\n" crate.name custom_msg;
+    List.iter (fun m -> Printf.fprintf out "import %s\n" m) imports;
+    close_out out;
+    log#linfo (lazy ("Generated: " ^ filename))
+  in
+
+  (* Emit one module per SCC, in topological (dependency-first) order; each
+     module is a chain of one or more opacity layers. *)
```
sonmarcho, line 1506: I think we need to clarify somewhere what a module is, what a layer is, etc. and how things are grouped/split into files

sonmarcho (commented):
Thanks for this PR: it's a lot of hard work!

I left a lot of comments, but at a high-level I have two:

- for AI assisted PRs, especially the ones that touch the translation (which is one of the most important parts of the TCB) I will be very suspicious of all details. This means the code has to be clean and extremely well documented. In the present state, comments look AI generated meaning they are either imprecise (AI jargon), useless (many things are documented that do not need to be documented), or missing (many functions or algorithms that would deserve to be explained in detail are not). Consequently there are many parts that I could not properly understand. I would love it if you could make a pass over the comments (I flagged a lot of them but not all of them - I think we should do a pass over all comments in the PR): I think rewriting them in our own (human) words is a good way of making sure we own the code and that nothing escaped our attention. My goal would be to understand what the code does just by reading the comments (I'm setting the bar quite high so it's ok if it takes time: I think it's very valuable).

- for the definitions that are generated by macros: I think the Charon spans do not only give us the span of the macro declaration but also the span where it is used (that is the span we want to use for the declaration), so we don't need to guess this span. Or at least I think there is a way of retrieving this information. Could you investigate? Maybe @Nadrieril can confirm.

Nadrieril:
Yep, each span have the "main" span, adn a `generated_from` span, which should give you the origin of the macro

Issue #1099, opened 2026-06-03
Fail to translate function with double reference and try (`?`)


niooss-ledger:
Hello,
While trying to extract Lean from some Rust code, I stumbled upon an Aeneas error when the code is using double references (`&&`) and `try` construction (operator `?`). For example, `func` from this code fails to be translated:

```rust
fn do_option() -> Option<()> {
    Some(())
}

pub fn func(_arg: &&u8) -> Option<()> {
    do_option()?;
    Some(())
}
```

```console
$ charon cargo --preset=aeneas && aeneas -print-error-emitters -backend lean mytest.llbc
[Info ] Imported: mytest.llbc
Applied prepasses:  [####################################################] 8/8 ✔️
Translated globals:  [---------------------------------------------------] 0/0 ✔️
Translated opaque functions:  [##########################################] 2/2 ✔️
[Error] Unimplementedt functions:  [-------------------------------------] 0/2 ⠋
Source: 'src/lib.rs', lines 6:4-8:1
Compiler source: interp/InterpJoin.ml, line 102
[Warn ] Could not translate the body of function 'mytest::func because of previous error
Name pattern: 'mytest::func'
Definition span: 'src/lib.rs', lines 5:0-8:1
Source: 'src/lib.rs', lines 6:4-8:1
Compiler source: interp/Interp.ml, line 550

Translated transparent functions:  [#####################################] 2/2 ✔️
Translated trait declarations:  [----------------------------------------] 0/3 ✔️
Translated trait impls:  [-----------------------------------------------] 0/3 ✔️
Post-processed translated opaque functions:  [###########################] 2/2 ✔️
Post-processed translated transparent functions:  [######################] 2/2 ✔️
[Info ] Generated the partial file (because of 1 errors, including 1 unique errors): ./Mytest.lean
[Warn ] The crate contains extracted external, unknown definitions: we advise using the option -split-files to allow manually providing these definitions in separate files.
[Info ] Total execution time: 0.154423 seconds
```

The same issue occurs when using `Result` instead of `Option`:

```rust
fn do_result() -> Result<(), ()> {
    Ok(())
}

pub fn func(_arg: &&u8) -> Result<(), ()> {
    do_result()?;
    Ok(())
}
```

Moreover:
- Using single references (`_arg: &u8`) does not trigger any error.
- Replacing `?` with `let _ = do_option();` also does not trigger any error.

[...]

PR #1189, opened 2026-07-06
Fix translation of nested shared borrow used in loop

niooss-ledger:
Hello,

After Pull Request #1187 fixed the simple reproducer from #1099, I tried to use Aeneas again on [some simple Rust code using nested shared borrow](https://github.com/LedgerHQ/app-sui/blob/e4176c8e0412da8417f4315799679246b87db7c2/rust-app/src/crypto_helpers/common.rs#L17-L28) and it failed. A minimal reproducer is:

```rust
pub struct RefSlice<'a> {
    pub s: &'a [u8],
}

pub fn xor_slice(rs: &RefSlice<'_>)  {
    for _ in rs.s {
    }
}
```

```text
$ aeneas -backend=lean -print-error-emitters mytest.llbc 
...
[Error] Internal error, please file an issue
Source: 'src/lib.rs', lines 6:4-7:5
Compiler source: interp/InterpMatchCtxs.ml, line 230
[Warn ] Could not translate the body of function 'mytest::xor_slice because of previous error
Name pattern: 'mytest::xor_slice'
Definition span: 'src/lib.rs', lines 5:0-8:1
Source: 'src/lib.rs', lines 6:4-7:5
Compiler source: interp/Interp.ml, line 609
```

The error occurs in https://github.com/AeneasVerif/aeneas/blob/bd30f6076f5d9a7f6c1db1b067e5aa43e7af4691/src/interp/InterpMatchCtxs.ml#L230

I wrote a test (in `tests/src/nested-shared-borrows.rs`) and let Claude Code with Sonnet 4.6 tell me what was going on and suggest me what was failing. It wrote a patch and explained:

> Root cause: PR #1187 introduced nested `VSharedBorrow` entries inside `ASharedLoan` shared values. This broke four code paths that assumed shared values had no borrows.
>
> Fix 1 — `explore_abs` visitor, `AProjSharedBorrow` case:
> Explicitly iterate `AsbBorrow` entries and call `register_borrow_id`, instead of calling `super` which triggered the `visit_borrow_id` internal error guard.
>
> Fix 2 — `explore_abs` visitor, `visit_borrow_content`:
> Added `method! visit_borrow_content _ _ = ()` to silence concrete borrows inside abstract values. The `ASharedLoan` shared value (which now contains `VSharedBorrow`) was triggering the `visit_borrow_id` `[%internal_error]` guard.
>
> Fix 3 — [I am removing this "fix" as it was actually some misunderstanding about what `phys_eq` does: the AI wanted to replace it with `not (equal_tvalue ...)`]
>
> Fix 4 — `ASharedLoan` matching sanity check:
> Changed `value_has_borrows sv.value` → `value_has_mut_borrows sv.value` in the `[%sanity_check_recover]` after matching a shared loan's concrete value. Shared values can now contain nested shared borrows (`VSharedBorrow`) by design, but still cannot contain mutable borrows.

I edited the changes to be comfortable with them (and I verified the test suite passes). Could you please review?

### DIFF
--- a/src/interp/InterpMatchCtxs.ml
+++ b/src/interp/InterpMatchCtxs.ml
@@ -215,10 +215,18 @@ let compute_abs_borrows_loans_maps (span : Meta.span) (explore : abs -> bool)
         | ASharedBorrow (npm, bid, sid) ->
             (* Add the current marker when visiting the borrow id *)
             register_borrow_id abs npm bid (Some sid)
-        | AProjSharedBorrow _ ->
+        | AProjSharedBorrow asb ->
             [%sanity_check] span (pm = PNone);
-            (* Process those normally *)
-            super#visit_aborrow_content (abs, pm) bc
+            (* Register the concrete shared borrows nested inside the projection.
+               We can't call super here because it would trigger visit_borrow_id
+               (which raises an internal error) for the AsbBorrow entries. *)
+            List.iter
+              (fun ab ->
+                match ab with
+                | AsbBorrow (bid, sid) ->
+                    register_borrow_id abs PNone bid (Some sid)
+                | AsbProjReborrows _ -> ())
+              asb
         | AIgnoredMutBorrow (_, child)
         | AEndedIgnoredMutBorrow { child; given_back = _; given_back_meta = _ }
           ->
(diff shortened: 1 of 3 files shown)


ojeda:
I have also hit this one, minimized to:

```rust
fn f(_v: &&u8) {
    for _ in 0..1 {}
}
```

It fails with:

```text
[Error] Internal error, please file an issue
Compiler source: interp/InterpMatchCtxs.ml, line 230
```

Replacing `&&u8` with `&u8` makes it translate.

This PR appears to fix it.

**Versions**:

- Aeneas: daa85d7e8940 (i.e. latest as of the time of submission)
- Charon: 0.1.232

PR #1208, opened 2026-07-14
Fix Usize inequality lifting in bvify

ia0:
This PR fixes a regression introduced by #1000 by adding the missing `Usize.lt_bv` and `Usize.le_bv` theorems for `bvify`.

Example proof hitting the regression: https://github.com/AeneasVerif/jxl-proofs/blob/f365cc22b372396949021d36c321f9033d0f326b/lean/JxlProofs/Proofs.lean#L86

Gemini was used to debug the issue and author the fix. I have reviewed the fix, confirmed that it fixes the issue, authored the commit, and created the PR.


sonmarcho:
Thanks!

PR #1213, opened 2026-07-16
CI: do not use self-hosted runner on forks

abentkamp:
On our fork cryspen/aeneas, we don't have self-hosted runners any more, so we need to turn them off. The changes made here affect only forks, but might be useful for other forks, too. It would be great to upstream this, since it would be one less commit that we need to rebase regularly.

This PR is entirely Claude generated, but has been carefully reviewed and tested on our fork.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,9 @@ on:
   workflow_dispatch:
 
 jobs:
+  select-runner:
+    uses: ./.github/workflows/select-runner.yml
+
   prepare:
     name: Prepare Release
     runs-on: ubuntu-latest
@@ -39,7 +42,7 @@ jobs:
 
   build:
     name: Build (${{ matrix.artifact_name }})
-    needs: prepare
+    needs: [prepare, select-runner]
     runs-on: ${{ matrix.os }}
     permissions:
       contents: write
@@ -48,10 +51,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: [self-hosted, linux, nix]
+          - os: ${{ fromJSON(needs.select-runner.outputs.runs_on) }}
             artifact_name: aeneas-linux-x86_64
             nix_attr: aeneas-static-release
-            nix_machine: true
+            nix_machine: ${{ contains(fromJSON(needs.select-runner.outputs.runs_on), 'nix') }}
           - os: ubuntu-24.04-arm
             artifact_name: aeneas-linux-aarch64
             nix_attr: aeneas-static-release
@@ -69,7 +72,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install nix
-        if: ${{ ! matrix.nix_machine }}
+        if: ${{ !matrix.nix_machine }}
         uses: nixbuild/nix-quick-install-action@v30
         with:
           nix_conf: |
@@ -79,7 +82,7 @@ jobs:
             extra-trusted-public-keys = hacl.cachix.org-1:FzsZ2xsByOwKwIWNPII7yMOelJNDZ12mDAj3d1eGX0c=
 
       - name: Restore Nix Cache
-        if: ${{ ! matrix.nix_machine }}
+        if: ${{ !matrix.nix_machine }}
         id: restore-nix-cache
         uses: nix-community/cache-nix-action/restore@v6
         with:
@@ -102,7 +105,7 @@ jobs:
           chmod -R +w dist_staging
 
       - name: Install elan (Lean toolchain manager)
-        if: ${{ ! matrix.nix_machine }}
+        if: ${{ !matrix.nix_machine }}
         run: |
           set -eo pipefail
 
@@ -171,7 +174,7 @@ jobs:
           files: ${{ matrix.artifact_name }}.tar.gz
 
       - name: Save Nix Cache
-        if: ${{ always() && ! matrix.nix_machine }}
+        if: ${{ always() && !matrix.nix_machine }}
         uses: nix-community/cache-nix-action/save@v6
         with:
           primary-key: ${{ steps.restore-nix-cache.outputs.primary-key }}
@@ -181,8 +184,8 @@ jobs:
 
   notify:
     name: Notify Zulip
-    needs: [prepare, build]
-    runs-on: ubuntu-latest
+    needs: [prepare, build, select-runner]
+    runs-on: ${{ fromJSON(needs.select-runner.outputs.runs_on) }}
     if: ${{ success() || failure() }}
     permissions:
       actions: read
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -40,36 +40,44 @@ jobs:
...
       matrix:
         include:
-          - os: [self-hosted, linux, nix]
+          # On the main repo this runs on the self-hosted runner; on forks it
+          # falls back to `os` (see `runs-on` and `is_self_hosted` above).
+          - os: ubuntu-24.04
             artifact_name: aeneas-linux-x86_64
             nix_attr: aeneas-static-release
-            nix_machine: true
+            self_hosted: true
```
Nadrieril: Why rename the variable? `nix_machine` is more accurate since someone could have a self-hosted setup that isn't nix-native

.github/workflows/release.yml
```diff
@@ -40,36 +40,44 @@ jobs:
   build:
     name: Build (${{ matrix.artifact_name }})
     needs: prepare
-    runs-on: ${{ matrix.os }}
+    # The self-hosted runner only exists on aeneasverif/aeneas. On forks (and
+    # any other repo), fall back to the GitHub-hosted `matrix.os` runner.
+    runs-on: ${{ (matrix.self_hosted && github.repository == 'aeneasverif/aeneas') && fromJSON('["self-hosted","linux","nix"]') || matrix.os }}
```
Nadrieril: Could you make the linux release job also use the select-runner logic instead of doing an ad-hoc thing here?

.github/workflows/release.yml
abentkamp: Ok, makes sense

.github/workflows/release.yml
abentkamp: Done

.github/workflows/release.yml
```diff
@@ -5,9 +5,13 @@ on:
...
 
 jobs:
+  select-runner:
+    uses: ./.github/workflows/select-runner.yml
+
   prepare:
     name: Prepare Release
-    runs-on: ubuntu-latest
+    needs: [select-runner]
+    runs-on: ${{ fromJSON(needs.select-runner.outputs.runs_on) }}
```
Nadrieril: This one does not benefit from self-hosted, just keep it on ubuntu-latest

.github/workflows/select-runner.yml
```diff
@@ -24,6 +26,14 @@ jobs:
           github_hosted='["ubuntu-latest"]'
           max_age_seconds=$((3 * 60 * 60))
 
+          # The self-hosted runner is only available on the main Aeneas repo.
+          # On forks (and any other repo), always fall back to GitHub-hosted.
+          if [ "$current_repository" != "aeneasverif/aeneas" ]; then
```
Nadrieril: shouldn't it be?
```suggestion
          if [ "$current_repository" != "AeneasVerif/aeneas" ]; then
```

Nadrieril: yes it should, I can see it on the current job

Nadrieril:
> This PR is entirely Claude generated, but has been carefully reviewed and tested on our fork.

The review comments I made all felt like I was thinking in your stead; I don't think a human would have made these choices. I do not enjoy that. When you review LLM work, please make the effort to imagine how you'd solve it and steer it to do that.

abentkamp:
> The review comments I made all felt like I was thinking in your stead; I don't think a human would have made these choices. I do not enjoy that. When you review LLM work, please make the effort to imagine how you'd solve it and steer it to do that.

Yes, I noticed that, too. Sorry, I should have been more careful. I'll have another look next week.

Nadrieril:
No worries, I know how easy that is, I've fallen for it many times.

Nadrieril:
Happy to merge the PR once you addressed the last comments I made btw

oliver-butterley:
Is this really the desired behaviour? I was under the impression that the release workflow running on cron should only run on the main repo, never on forks. 

It is a bit inconvenient that on all forks the cron errors each day when it tries to run.

How about we instead gate the entire release workflow so it only runs on the main repo, not on forks?

Nadrieril:
> It is a bit inconvenient that on all forks the cron errors each day when it tries to run.

Why does it error? A fork would be able to make their own releases too, no?

I honestly don't know what's the standard for cases like these. CI is obviously kinda repo-specific :/

oliver-butterley:
> > It is a bit inconvenient that on all forks the cron errors each day when it tries to run.
> 
> Why does it error? A fork would be able to make their own releases too, no?
> 
> I honestly don't know what's the standard for cases like these. CI is obviously kinda repo-specific :/

Sure, forks can make their own releases. I just meant to say that in most cases this isn't the desired behaviour. For myself and for several others I know, we have forks of Aeneas in order to contribute PRs to Aeneas. As such making releases for the fork isn't useful and is just clutter. On the other hand this is a minor quality of life thing since we can always manually deactivate the workflow for our repos. 

The error I mentioned was simply that the release workflow typically gives an error or gets cancelled due to waiting too long. Although I never investigated why since I didn't want the workflow running.

The way that `release.yml` automatically generates a github release and includes a step for posting to Zulip really strongly suggests that this is a workflow designed for the main repo.

Nadrieril:
> As such making releases for the fork isn't useful and is just clutter.

If you have a fork only for contributing, just disable github actions on it entirely, no?

> The way that release.yml automatically generates a github release and includes a step for posting to Zulip really strongly suggests that this is a workflow designed for the main repo.

Oh I forgot about the Zulip posting, yeah ok.

@abentkamp do you have a use for automated releases on your fork?

abentkamp:
Yes, we are currently relying on the releases of our fork. The intention is to move everything upstream, but having the fork releases allows us to make faster progress.

Limiting the Zulip posting to AeneasVerif/aeneas only would make sense, though. That one is also failing on our fork.

If we really want to disable releases on forks by default, we could add a GitHub environment variable `RELEASE` that must be set for the release job to run. Then we could add that environment variable only to upstream and to the forks that want releases. What do you think?

.github/workflows/release.yml
abentkamp: Okay, so self-hosted is only for jobs using nix?

.github/workflows/select-runner.yml
abentkamp: Okay, done.

abentkamp:
I've addressed @Nadrieril's comments and limited the Zulip post to the main repo.

Nadrieril:
Nice yeah, I like this!

Nadrieril:
could you just rebase/merge with latest main? Then I can merge

abentkamp:
Done. I also squashed some of the commits.

PR #1224, opened 2026-07-20
Minimize the generic parameters of globals/constants

### LINKED ISSUE #336: Minimize the parameters of globals/constants
If a global constant is defined in an implementation, we parameterize it with all the generics of the impl block (see [here](https://github.com/AeneasVerif/aeneas/blob/a5f5bf6dcfa62a4c1eefa339edd424cbfb433ec4/tests/lean/Traits.lean#L543) for instance). We should only parameterize it with the generics it actually uses. Note that doing this should only require implementing a simple micro-pass.

oliver-butterley:
Previously a global/const declared in a generic impl block (or a defaulted trait const) inherited all the generics of the enclosing impl/trait. It now minimizes the parameters when it can. 

Implemented as a crate-level micro-pass (`PureMicroPasses.minimize_globals_generics`) which computes, per global, the parameters used by its type and initializer body, shrinks the declaration (generics + recomputed explicit_info), and filters the generic arguments at every use site (Global qualifiers in bodies, trait-impl const refs).

Builtin globals and opaque globals are skipped.

Adds `tests/src/issue-336-global-generics.rs` covering multiple different instances of this.

Fixes #336


Nadrieril:
> opaque globals are minimized from their type alone.

That can be a problem, e.g. in a case like
```rust
const SIZE<T>: usize = size_of::<T>():
```

oliver-butterley:
> > opaque globals are minimized from their type alone.
> 
> That can be a problem, e.g. in a case like
> 
> ```rust
> const SIZE<T>: usize = size_of::<T>():
> ```

Ah! Thanks for the example. Indeed, the PR is still in draft status for very good reason!

oliver-butterley:
@Nadrieril I fixed the outstanding problems, like the one you noticed. So possibly this code is now good. However I will spend some more time making sure that everything is fine before marking it ready for review.

Issue #1225, opened 2026-07-20
Merge the `Error` variants


sonmarcho:
Today `Error` (used by `Result.fail`) has quite a few variants. They should be merged into `panic` (for panicking executions) and `ub` (for executions that trigger UB): this would allow reasoning about panicking executions, while enforcing that we always need to prove that an execution doesn't trigger UB.

protz:
This means that we need to tweak the signature of the `unchecked` operations like `get_unchecked` to reflect that failure to be in bounds triggers UB, not panic.

joshlf:
Slight preference for `ub` being top-level (i.e., `Result.ub`) since encountering UB means *any* execution is possible, including running forever. By contrast, the current `div`/`fail` split implies that `fail` cases always halt execution.

If we want to be *really* pedantic (which we don't - I'm just presenting this to explain my mental model), the way I model this in my head is:

```lean4
inductive Error where
   ...

inductive WellDefResult (α : Type u) where
  | ok (v: α): WellDefResult α
  | fail (e: Error): WellDefResult α
  | div

inductive Result (α : Type u) where
  | defined (r: WellDefResult α)
  | ub
```

In practice, I think it's fine (and probably more ergonomic) to do:


```lean4
inductive Error where
   ...

inductive Result (α : Type u) where
  | ok (v: α): Result α
  | fail (e: Error): Result α
  | div
  | ub
```

mdittmer:
> This means that we need to tweak the signature of the `unchecked` operations like `get_unchecked` to reflect that failure to be in bounds triggers UB, not panic.

@protz I had naively assumed that all possible source of UB appear as opaque holes in Aeneas's transformation. Is that not the case? Do we have a sense of "how many of the sources of UB" show up as a transparent Aeneas transformation?

joshlf:
> > This means that we need to tweak the signature of the `unchecked` operations like `get_unchecked` to reflect that failure to be in bounds triggers UB, not panic.
> 
> [@protz](https://github.com/protz) I had naively assumed that all possible source of UB appear as opaque holes in Aeneas's transformation. Is that not the case? Do we have a sense of "how many of the sources of UB" show up as a transparent Aeneas transformation?

[...]

PR #1226, opened 2026-07-22
Introduce `ub` variant for `Result`

mdittmer:
This change introduces a `Result` variant intended to model undefined behaviour (UB). The intent is to allow unsafe definitions and theorems to introduce proof obligations that formalize the conditions under which unsafe operations introduce undefined behaviour. Using a `Result` variant provides a natural abstraction asserting that undefined behaviour at a particular step determines all subsequent behaviour as undefined.

Towards #1225

AI assisted this PR as my "Lean tutor", helping me know where to look and what abstractions to reach for but not authoring any code.

### DIFF
--- a/backends/lean/Aeneas/Std/Array/Core.lean
+++ b/backends/lean/Aeneas/Std/Array/Core.lean
@@ -37,6 +37,7 @@ def List.clone (clone : α → Result α) (l : List α) : Result ({ l' : List α
   match h :List.mapM clone l with
   | ok v => ok ⟨ v, by have := List.mapM_Result_length h; scalar_tac ⟩
   | fail e => fail e
+  | ub => ub
   | div => div
 
 @[step]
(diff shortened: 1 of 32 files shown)


sonmarcho:
Thanks for this! What I had in mind was not to modify `Result` but `Error` (`UB` is one kind of error).

mdittmer:
Yeah. I would appreciate more discussion about this question. I tend to agree with @joshlf 's [assessment](https://github.com/AeneasVerif/aeneas/issues/1225#issuecomment-5025619318). The point that I would emphasize is that, AFAICT, developers can quite reasonably assume that `Result.fail` implies (a certain kind of) termination. The curious thing about UB, as I understand it, is that all bets are off: in the presence of UB you cannot be confident that a program will terminate or diverge; in fact, you cannot be confident in much of anything!

joshlf:
> in fact, you cannot be confident in much of anything!

[`Result.nasaldemons`](http://catb.org/jargon/html/N/nasal-demons.html)?

PR #1230, opened 2026-07-24
 Migrate Aeneas to the Lean module system

oliver-butterley:
The PR is twofold:

- Migrates the Aeneas backend lean project to the module system
- Updates the translation engine to produce Lean files which use the module system

This PR does the migration in the way that there is minimal change to the code and minimal effect to downstream users. As such, in general the full benefits of the module system aren't being used, instead many definitions are `@[expose]`d in order to maintain the downstream proofs which took advantage of the visibility into all defs. This was done methodically in a way that something was only ever marked `meta` if it was forced and something was only every marked `expose` if that was required to keep downstream behaviour unchanged.


oliver-butterley:
Note: The code produced by `gen-lib-entry` has been updated to support the module system but trying to implement a test for this revealed there was already a bug in the placement of files making the test not practical. Locally tested to do the correct thing, modulo the known bug.

See Zulip discussion: [#general > ✔ -gen-lib-entry creates all files at the root](https://aeneas-verif.zulipchat.com/#narrow/channel/349819-general/topic/.E2.9C.94.20-gen-lib-entry.20creates.20all.20files.20at.20the.20root/with/613146578).

backends/lean/Aeneas/Std/Scalar/Core.lean
```diff
@@ -666,20 +672,20 @@ abbrev  I128  := IScalar .I128
 
 /-!  ofNatCore -/
 -- TODO: typeclass?
-def Usize.ofNatCore := @UScalar.ofNatCore .Usize
-def U8.ofNatCore    := @UScalar.ofNatCore .U8
-def U16.ofNatCore   := @UScalar.ofNatCore .U16
-def U32.ofNatCore   := @UScalar.ofNatCore .U32
-def U64.ofNatCore   := @UScalar.ofNatCore .U64
-def U128.ofNatCore  := @UScalar.ofNatCore .U128
+@[expose] def Usize.ofNatCore := @UScalar.ofNatCore .Usize
```
jeprinz: @[expose] section could be nice here

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -1,30 +1,33 @@
...
+public meta import Aeneas.Std.Delab
+public section
 
 namespace Aeneas.Std.WP
 
 open Std Result
 
-def Post α := (α -> Prop)
-def Pre := Prop
+abbrev Post α := (α -> Prop)
```
jeprinz:
```suggestion
@[expose] def Post α := (α -> Prop)
```

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -1,30 +1,33 @@
...
+abbrev Pre := Prop
```
jeprinz:
```suggestion
@[expose] def Pre := Prop
```

backends/lean/Aeneas/Std/WP.lean
```diff
@@ -1,30 +1,33 @@
...
 
-def Wp α := Post α → Pre
+abbrev Wp α := Post α → Pre
```
jeprinz:
```suggestion
@[expose] def Wp α := Post α → Pre
```

backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean
```diff
@@ -318,6 +322,9 @@ def resolveStructFields (value : Expr) (selfFvarId : FVarId) (type : Expr) : Met
        compiles, and applies any declaration attributes (e.g., `@[simp]`). -/
 elab mods:declModifiers "impl_def " id:declId sig:optDeclSig val:declVal : command => do
   let modifiers ← elabModifiers mods
+  let modifiers :=
+    if modifiers.visibility matches .regular then { modifiers with visibility := .public }
```
jeprinz: This should ideally use the visibility from the surrounding scope rather than assuming it is .public

jeprinz (approved):
I had some small nitpicks.

About the #assert issue, what is currently requiring assertions to compile the code is that computation of the tests depends on `System.Platform.numBits`, which is used in various places in aeneas and only computes when the code is compiled.

If there are any rust files that have both definitions and tests in them then maybe when `use-lean-modules true` it should just skip the tests?

backends/lean/Aeneas/Std/Scalar/Core.lean
oliver-butterley: good point! done. there might be other places like this... I will look

oliver-butterley:
> About the #assert issue, what is currently requiring assertions to compile the code is that computation of the tests depends on `System.Platform.numBits`, which is used in various places in aeneas and only computes when the code is compiled.

I'm troubled by the use of `System.Platform.numBits` in multiple ways! I don't see why being 32/64 on the machine where the Lean code is running is relevant to the verification work. The Lean code is meant to be a faithful model of the Rust code, there is no reason why we shouldn't be modelling a 32 bit machine on a 64 bit machine, etc. Although in reality we never see proofs where it matters if it is 32 or 64. Maybe I'm missing something but I would prefer numBits used in the verification to be something that can be set to 32 or 64 or typically verification doesn't use anything about it except that it is either 32 or 64 and that is all that is used in any proof. Not sure the cleanest way to do this. Perhaps an opaque with a axiom that it is either 32 or 64 but with a mechanism that it can be set to one specific value if desired during verification work. 

> If there are any rust files that have both definitions and tests in them then maybe when `use-lean-modules true` it should just skip the tests?

The current situation is that the docstring of #assert tells the user about the limitation in combination with the module system. And in the tests each one that requires #assert has the manually specified Aeneas args to avoid the module system. Another option would be to use precompiled modules with Aeneas but that seems very heavy in general so that should be a config option on import. If I understood correctly the suggestion, I'm not completely in favour because it can easily mean that a test is silently ignored even though it was added with the intent to check something important. The general situation isn't completely comfortable. Is there a way we can modify #assert so it works generally?

oliver-butterley:
Thanks @jeprinz for all the feedback. I made the modifications you suggested. I'm now wondering if there are a few other places which should be exposed and I missed them. I tried to be as conservative as possible with this but perhaps it was a greater priority to make sure everything remains unchanged downstream and then later we will fully embrace the module system in Aeneas, file by file. I'll take another careful look. EDIT: I updated a few other places which would be better as meta.

backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean
oliver-butterley: Yes, I agree with the desired functionality. I might have updated it correctly but I would need to double check this modification.

oliver-butterley:
TODO: 
- Investigate the assert in tests issue. (EDIT: done)

backends/lean/Aeneas/Command/Decompose/Tests.lean
```diff
@@ -1,10 +1,12 @@
...
-import Aeneas.Command.Decompose
-import Aeneas.Std
-import Aeneas.Do.Elab
-import Aeneas.Do.Delab
+module
+public import Aeneas.Command.Decompose
+public import Aeneas.Std
+public import Aeneas.Do.Elab
+public import Aeneas.Do.Delab
+public section
```
sonmarcho, line 9: Does it really make sense to mark tests as `public section` (also see the other files in the `Decompose` folder, `Step/Tests/`, etc.)? I think tests should be kept private?

backends/lean/Aeneas/Do/Tests.lean
```diff
@@ -1,6 +1,8 @@
-import Aeneas.Do.Elab
-import Aeneas.Do.Delab
-import Aeneas.Std
+module
+public import Aeneas.Do.Elab
+public import Aeneas.Do.Delab
+public import Aeneas.Std
+public section
```
sonmarcho, line 5: Same remark as for the `Decompose` tests

backends/lean/Aeneas/Std/Array/ArraySlice.lean
```diff
@@ -25,6 +30,8 @@ def Array.from_slice {α : Type u} {n : Usize} (a : Array α n) (s : Slice α) :
     .from s.val (by simp [*])
   else a -- Unreachable case
 
+end
```
sonmarcho, line 33: Having a section here is probably overkill: better individually mark the definitions as `expose`

backends/lean/Aeneas/Std/Scalar/CloneCopy.lean
```diff
@@ -10,16 +12,16 @@ open Result Error ScalarElab
 -/
 
 -- Remark: the command `uscalar` turns the name `Clone'S` into `CloneU8`, `CloneU16`, etc.
-uscalar @[reducible, simp] def core.clone.impls.Clone'S.clone (x : «%S») : «%S» := x
+uscalar @[reducible, simp, expose] def core.clone.impls.Clone'S.clone (x : «%S») : «%S» := x
```
sonmarcho, line 15: Generally speaking would it be possible to put the `expose` attribute first when there are several attributes  (I think it makes sense to put visibility first)?

backends/lean/Aeneas/Tactic/Conv/Bvify/Bvify.lean
```diff
@@ -23,17 +25,48 @@ structure Config where
...
+dsimproc [bvify] Int.reduceDiv' ((_ / _ : Int)) := Int.reduceDiv
+dsimproc [bvify] Int.reduceMod' ((_ % _ : Int)) := Int.reduceMod
+dsimproc [bvify] Int.reduceNegSucc' (Int.negSucc _) := Int.reduceNegSucc
+dsimproc [bvify] Int.reduceNeg' ((- _ : Int)) := Int.reduceNeg
+dsimproc [bvify] Int.reduceToNat' (Int.toNat _) := Int.reduceToNat
+dsimproc [bvify] BitVec.reduceMul' ((_ * _ : BitVec _)) := BitVec.reduceMul
+dsimproc [bvify] BitVec.reduceAdd' ((_ + _ : BitVec _)) := BitVec.reduceAdd
+dsimproc [bvify] BitVec.reduceSub' ((_ - _ : BitVec _)) := BitVec.reduceSub
+dsimproc [bvify] BitVec.reduceMod' ((_ % _ : BitVec _)) := BitVec.reduceMod
+dsimproc [bvify] BitVec.reduceDiv' ((_ / _ : BitVec _)) := BitVec.reduceDiv
```
sonmarcho, line 68: What's the issue here?

backends/lean/Aeneas/Tactic/Conv/ZModify/ZModify.lean
```diff
@@ -102,11 +106,10 @@ elab stx:zmodify : tactic =>
   zmodifyTac config n args loc
 
 attribute [zmodify] Nat.eq_mod_iff_eq_ZMod Int.eq_mod_iff_eq_ZMod div_to_ZMod
-attribute [zmodify] Nat.reduceGcd
-
+dsimproc [zmodify] Nat.reduceGcd' (Nat.gcd _ _) := Nat.reduceGcd
 attribute [zmodify] ZMod.natCast_val ZMod.natCast_mod ZMod.cast_id' ZMod.intCast_mod id_eq
 attribute [zmodify] Int.cast_add Int.cast_natCast Int.cast_mul
-attribute [zmodify] Int.reduceNeg
+dsimproc [zmodify] Int.reduceNeg' ((- _ : Int)) := Int.reduceNeg
```
sonmarcho, line 112: Similar question as above: why doesn't using an attribute work anymore?

backends/lean/Aeneas/Tactic/Simp/SimpBoolProp/SimpBoolProp.lean
```diff
@@ -9,8 +12,10 @@ namespace Aeneas.SimpBoolProp
 
 open Lean Lean.Meta Lean.Parser.Tactic Lean.Elab.Tactic
 
+/-- TODO: builtin simprocs can't be directly added under the module system. -/
+simproc ↓ [simp_bool_prop] reduceIte' (ite _ _ _) := reduceIte
```
sonmarcho, line 16: Similar as above

backends/lean/Aeneas/Tactic/Solver/BvTac/Tests.lean
```diff
@@ -1,5 +1,7 @@
-import Aeneas.Tactic.Solver.BvTac.BvTac
-import Aeneas.Std
+module
+public import Aeneas.Tactic.Solver.BvTac.BvTac
+public import Aeneas.Std
+public section
```
sonmarcho, line 4: Similar remark as for the other test files (I will not comment them all)

backends/lean/Aeneas/Tactic/Step/Tests/Coin.lean
```diff
@@ -1,5 +1,8 @@
-import Aeneas.Std
-import Aeneas.Tactic.Step
+module
+public import Aeneas.Std
+public import Aeneas.Tactic.Step
+
+@[expose] public section
```
sonmarcho, line 5: This is just a test, so it should be private and not exposed

backends/lean/AeneasMeta/Extensions.lean
```diff
@@ -26,7 +28,7 @@ def mkListDeclarationExtension [Inhabited α] (name : Name := by exact decl_name
     toArrayFn     := fun l => l.toArray
   }
 
-def SetDeclarationExtension := SimplePersistentEnvExtension Name NameSet
+@[expose] def SetDeclarationExtension := SimplePersistentEnvExtension Name NameSet
```
sonmarcho, line 31: I think these should only be meta?

src/Main.ml
```diff
@@ -105,6 +105,10 @@ let () =
         Arg.String set_subdir,
         " Extract the files in a sub-folder; this option has an impact on the \
          import paths of the generated files" );
+      ( "-use-lean-modules",
+        Arg.Bool (fun b -> use_lean_modules := b),
+        " Emit Lean files using the module system (default: true). Pass \
+         `-use-lean-modules false` to emit non-module files." );
```
sonmarcho, line 111: "to emit files that don't use the module system"?

tests/lean/Avl/Funs.lean
```diff
@@ -1,7 +1,9 @@
+module
```
sonmarcho, line 1: Could we put the `module` below the `-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS` comment?

tests/src/chunks_exact.rs
```diff
@@ -1,3 +1,4 @@
+//@ [lean] aeneas-args=-use-lean-modules false
```
sonmarcho, line 1: Why this (and similar question for the other tests that receive `aeneas-args=-use-lean-modules false`)?

sonmarcho:
Thanks for this PR! I left a few minor comments: I'll merge it as soon as they are addressed.

backends/lean/Aeneas/Command/Decompose/Tests.lean
oliver-butterley: Yes, private is correct for them! Just an oversight, I'll correct all the test files.

backends/lean/Aeneas/Std/Array/ArraySlice.lean
oliver-butterley: Sure!

backends/lean/Aeneas/Std/Scalar/CloneCopy.lean
oliver-butterley: OK! I'll put them first in each case

backends/lean/Aeneas/Tactic/Conv/Bvify/Bvify.lean
oliver-butterley: Under the module system, it's not possible to add attributes to builtin simprocs. I'll investigate exactly why...

backends/lean/AeneasMeta/Extensions.lean
oliver-butterley: That seems likely. I'll check... that looks like I intentionally added it so I should figure out what my logic was.

src/Main.ml
oliver-butterley: sure, that's better phrasing

tests/lean/Avl/Funs.lean
oliver-butterley: absolutely, I'll change it

tests/src/chunks_exact.rs
oliver-butterley: To make `#assert` work in these cases I know of two ways: either deactivate the module system in those files or we need to add a meta import to the files which contain the `#assert`. 

Probably your next question is if there is another way! I wonder if it is possible to arrange Aeneas so the stuff needed by `#assert` is automatically available both in normal and meta.

backends/lean/AeneasMeta/Extensions.lean
```diff
@@ -12,7 +14,7 @@ namespace Extensions
 open Lean Elab Term Meta
 open Utils
 
-def ListDeclarationExtension (α : Type) := SimplePersistentEnvExtension α (List α)
+@[expose] def ListDeclarationExtension (α : Type) := SimplePersistentEnvExtension α (List α)
```
oliver-butterley, line 19: This expose is required for the following def and instance

backends/lean/AeneasMeta/Extensions.lean
oliver-butterley: No, both of these are required. I now remember adding them! If not, then some of the following def in this same file have the error:
```
Compilation failed, locally inferred compilation type differs from type that would be inferred in other modules. Some of the following definitions may need to be `@[expose]`d to fix this mismatch: 
  SetDeclarationExtension ↦ 1
This is a current compiler limitation for `module`s that may be lifted in the future.
```

backends/lean/Aeneas/Std/Primitives.lean
```diff
@@ -289,6 +300,11 @@ def Result.ofOption {a : Type u} (x : Option a) (e : Error) : Result a :=
   (Bind.bind (Bind.bind e g) h) =
   (Bind.bind e (λ x => Bind.bind (g x) h)) := by apply bind_assoc
 
+end
+
+unseal Result
```
sonmarcho, line 305: Why this?

backends/lean/Aeneas/Tactic/Step/Tests/Coin.lean
```diff
@@ -1,6 +1,8 @@
+module
 import Aeneas.Std
 import Aeneas.Tactic.Step
 
+
 /-! # Coinductive specifications over `ITree`s
```
sonmarcho:
```suggestion
/-! # Coinductive specifications over `ITree`s
```

backends/lean/Aeneas/Tactic/Step/Tests/Coin.lean
```diff
@@ -160,3 +162,4 @@ noncomputable instance {T} : Lean.Order.CCPO (ITreeC T) := instCCPOCoIndOfInhabi
 instance : MonoBind ITreeC := instMonoBindITree
 
 end Aeneas.Tactic.Step.Tests.Coin
+
```
sonmarcho, line 164:
```suggestion
end Aeneas.Tactic.Step.Tests.Coin
```

sonmarcho (approved):

PR #1235, opened 2026-07-27
fix: replace `sorry`s with `noncomputable opaque`/`axiom`

### LINKED ISSUE #18: Remove the sorries from the Lean backend
Some proofs in the Lean backend (`Primitives.lean`) are unfinished and use `sorry`

abentkamp:
This PR replaces all `sorry`s in the Lean backend library with `noncomputable opaque`/`axiom`. The advantage is that building the library no longer yields `declaration uses sorry`-warnings. Other than that, the library should work as before.

Only for `core.str.iter.IteratorChars.next`, giving a definition was straight-forward, so I defined it.

Fixes #18

### DIFF
--- a/backends/lean/Aeneas/Std/Slice.lean
+++ b/backends/lean/Aeneas/Std/Slice.lean
@@ -360,12 +360,11 @@ def core.slice.Slice.get
   inst.get i s
 
 @[rust_fun "core::slice::{[@T]}::get_unchecked"]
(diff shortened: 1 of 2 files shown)


clementblaudeau (approved):
The defs being in `Result`, using an opaque definition rather than a sorried one seems a good idea. Hopefully they'll eventually become real definitions.

PR #1236, opened 2026-07-28
add `spec_refl`

oliver-butterley:
`spec_refl` strengthens any spec's postcondition with the call identity `m = ok r`. The `refl_of%` elaborator lifts this over the binders of a `∀`-quantified spec theorem, so a reflexive spec can be dropped into a proof's local context where `step` picks them up.


oliver-butterley:
TODO: move the tests into a separate file which also imports more of Aeneas and so can use `step*` in a realistic example.

PR #1237, opened 2026-07-28
add api for slice/vec/array

oliver-butterley:
This PR adds various basic results for slice/vec/array that were not yet present in Aeneas but fit the pattern of what is included there and which help to make smooth proofs of spec theorems downstream.


oliver-butterley:
TODO: `Array.map` deserves to be included in these additions. 

```lean
/-- Map a function over an `Array`, preserving the length. -/
def Array.map {α : Type u} {β : Type v} {n : Usize} (f : α → β) (a : Array α n) : Array β n :=
  ⟨a.val.map f, by simp [a.property]⟩
```

EDIT: this def and two lemmas added

backends/lean/Aeneas/Std/Array/Array.lean
```diff
@@ -61,6 +61,22 @@ example : Result (Array Int (Usize.ofNat 2)) := do
   let y ← ok 1
   ok (Array.make (Usize.ofNat 2) [x, y])
 
+@[simp, grind =]
+theorem Array.val_make {α : Type u} (n : Usize) (l : List α) (h : l.length = n) :
+    (Array.make n l h).val = l := rfl
+
+@[scalar_tac Array.make n l h, grind =]
```
sonmarcho, line 68: We should add `simp`

backends/lean/Aeneas/Std/Array/Array.lean
```diff
@@ -61,6 +61,22 @@ example : Result (Array Int (Usize.ofNat 2)) := do
...
+theorem Array.length_make {α : Type u} (n : Usize) (l : List α) (h : l.length = n) :
+    (Array.make n l h).length = n.val := h
+
+@[simp, grind =]
+theorem Array.make_val {α : Type u} {n : Usize} (a : Array α n) (h : a.length = n) :
+    Array.make n a.val h = a := rfl
+
+theorem Array.make_inj {α : Type u} {n : Usize} (l₁ l₂ : List α) (h₁ : l₁.length = n)
+    (h₂ : l₂.length = n) : Array.make n l₁ h₁ = Array.make n l₂ h₂ ↔ l₁ = l₂ :=
```
sonmarcho, line 77: I think grind has an attribute for this kind of injectivity lemmas

oliver-butterley:
I suspect that this PR is redundant since #1256. Is that correct, @jeprinz? Then I'll go ahead and close it. Alternatively, is it useful for me to add the corresponding add api for slice/vec/array but now based on the new `ListN` foundations?

PR #1244, opened 2026-07-31
Update charon

Nadrieril:
Companion PR to https://github.com/AeneasVerif/charon/pull/1341

### DIFF
--- a/tests/lean/Arrays.lean
+++ b/tests/lean/Arrays.lean
@@ -197,6 +197,7 @@ def update_update_array
   :
   Result (Array (Array Std.U32 32#usize) 32#usize)
   := do
+  let _ ← Array.index_usize s i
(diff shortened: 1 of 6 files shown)


tests/lean/Arrays.lean
```diff
@@ -197,6 +197,7 @@ def update_update_array
   :
   Result (Array (Array Std.U32 32#usize) 32#usize)
   := do
+  let _ ← Array.index_usize s i
```
Nadrieril, line 200: This appears because there's a FakeRead of the place that we used to not translate and now we do. It's not the FakeRead itself that causes this, but Charon's attempt at turning the index place projection into a function call, as it usually does for aeneas. I didn't want to special-case this in charon but if that's a bother, I can.

PR #1254, opened 2026-08-11
Fix a bug with the namespacing of external items for the emit-json feature

### LINKED ISSUE #1248: Bug: `translation.json` `lean_name` for opaque functions carries the crate namespace prefix, but the emitted declaration in `FunsExternal_Template.lean` does not

```rust
// mycrate/src/lib.rs
pub mod ffi {
    pub fn opaque_fn(x: u32) -> u32 {
        x + 1
    }
}

pub fn caller(x: u32) -> u32 {
    ffi::opaque_fn(x)
}
```

`opaque_fn` is made opaque from the Charon side (`--opaque`), which is how the real project marks the functions it does not want translated. Any way of producing an `is_opaque: true` function should do.

**Charon command**: `charon cargo --preset=aeneas --opaque 'mycrate::ffi::opaque_fn'`

**Aeneas command**: `aeneas -backend lean -split-files -emit-json mycrate.llbc -dest .`

**Versions**:
- Aeneas: daa85d7e (main), prebuilt `nightly-2026.08.07-daa85d7`
- Charon: 0.1.232 (prebuilt `nightly-2026.08.07`)

**Aeneas output**:

```text
(no warnings or errors)
```

**Lean translation**:

Aeneas emits, for a crate split with `-split-files`:

- `Funs.lean` / `Types.lean`: translated declarations, wrapped in  `namespace <crate>` … `end <crate>`, so a def written `def ffi.other` there has the fully-qualified Lean name `mycrate.ffi.other`.
- `FunsExternal_Template.lean`: declarations for **opaque** functions (`is_opaque: true`). This file is *not* wrapped in `namespace <crate>`; it uses `open <crate>` instead. A declaration written `axiom ffi.opaque_fn` there has the bare Lean name `ffi.opaque_fn`, no `mycrate.` prefix.
- `TypesExternal_Template.lean`: opaque types. This file has neither `namespace <crate>` nor `open <crate>`, so its declarations are likewise bare.

`translation.json`, however, records `lean_name` for the declarations in both `*External_Template.lean` files **with** the crate prefix, i.e. the name they *would* have had under `namespace <crate>`, which is not how they were emitted.

*What is produced.* For the minimal crate above, `FunsExternal_Template.lean` is generated as (verbatim):

```lean
-- THIS FILE WAS AUTOMATICALLY GENERATED BY AENEAS
-- [mycrate]: external functions.
-- This is a template file: rename it to "FunsExternal.lean" and fill the holes.
import Aeneas
import Mycrate.Types
open Aeneas Aeneas.Std Result ControlFlow Error
...
open mycrate

/-- [mycrate::ffi::opaque_fn]:
    Source: 'src/lib.rs', lines 2:4-4:5
    Visibility: public -/
axiom ffi.opaque_fn : Std.U32 → Result Std.U32
```

so the constant is `ffi.opaque_fn`, whereas `Funs.lean` for the same crate does open `namespace mycrate` and contains `def caller`, i.e. `mycrate.caller`. The `translation.json` written by that same run records (verbatim, both entries):

```json
{
  "def_id": 1,
  "lean_name": "mycrate.ffi.opaque_fn",
  "lean_file": "FunsExternal_Template.lean",
  "rust_name": "mycrate::ffi::opaque_fn",
  "source": { "file": "src/lib.rs", "begin_line": 2, "end_line": 4 },
  "is_opaque": true
}
{
  "def_id": 0,
  "lean_name": "mycrate.caller",
  "lean_file": "Funs.lean",
  "rust_name": "mycrate::caller",
  "source": { "file": "src/lib.rs", "begin_line": 7, "end_line": 9 },
  "is_opaque": false
}
```

`mycrate.caller` is correct. `mycrate.ffi.opaque_fn` is not a constant that exists anywhere: the declaration emitted for it is `ffi.opaque_fn`.

The same happens on a real project. `translation.json`, `def_id` 509:

```json
{
  "def_id": 509,
  "lean_name": "spqr.encoding.polynomial.PolyDecoder.Insts.SpqrEncodingDecoder.decoded_message",
  "lean_file": "SrcTranslated/FunsExternal_Template.lean",
  "rust_name": "spqr::encoding::polynomial::{impl spqr::encoding::Decoder for spqr::encoding::polynomial::PolyDecoder}::decoded_message",
  "source": { "file": "src/encoding/polynomial.rs", "begin_line": 911, "end_line": 963 },
  "is_opaque": true
}
```

and the declaration actually emitted for it (kept in `SrcTranslated/FunsExternal.lean`, under `open spqr`, not `namespace spqr`):

```lean
axiom encoding.polynomial.PolyDecoder.Insts.SpqrEncodingDecoder.decoded_message
  : encoding.polynomial.PolyDecoder → Result (Option (alloc.vec.Vec Std.U8))
```

The manifest says `spqr.encoding.…decoded_message`; the real Lean constant is `encoding.…decoded_message`. In the Lean atoms extracted from the built project the prefixed name has **0** occurrences; the bare name is present.

This affects every declaration recorded against a `*External_Template.lean` file. Non-opaque declarations are fine: `spqr.encoding.polynomial.PolyDecoder.Insts.SpqrEncodingDecoder.add_chunk` in the manifest matches its `Funs.lean` constant exactly.

*What should instead be produced*: `lean_name` should equal the emitted Lean constant name: `mycrate.ffi.opaque_fn` should be `ffi.opaque_fn` (or the template files should be wrapped in `namespace <crate>` so the prefixed name becomes the true one; see below).

**Explain the bug**: (all links below point at `daa85d7e`)

[`extract_file_info`](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L1280-L1294) carries `namespace` alongside two independent flags, `in_namespace` and `open_namespace`, and [`extract_file` honours them when writing the header](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L1403-L1406):

```ocaml
(* src/Translate.ml:1404 *)
if fi.in_namespace then Printf.fprintf out "\nnamespace %s\n" fi.namespace;
if fi.open_namespace then Printf.fprintf out "open %s\n" fi.namespace
```

But [the manifest hook a hundred lines earlier](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L1302-L1303) is given the namespace unconditionally, with no reference to `in_namespace`:

```ocaml
(* src/Translate.ml:1303 *)
EmitJson.begin_file_if_enabled ~filename:fi.filename ~namespace:fi.namespace;
```

That namespace is [stashed in the emitter's state](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/EmitJson.ml#L273-L284), and [`EmitJson.full_lean_name`](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/EmitJson.ml#L152-L155) then prefixes every recorded short name with it, including [the `lean_name` of each function entry](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/EmitJson.ml#L169-L172). The two external template files are constructed with `in_namespace = false`, [the types one](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L1914-L1919) (also `open_namespace = false`) and [the funs one](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L2037-L2042) (`open_namespace = true`), so their declarations are not inside the crate namespace, yet the manifest names them as if they were.

The consequence is that a consumer that joins the manifest to the emitted Lean by `lean_name` misses on exactly the opaque declarations, and misses silently, the name is well-formed, it just does not exist.

A possible fix is to make the manifest match the emitted Lean: pass  `~namespace:(if fi.in_namespace then fi.namespace else "")` at [`src/Translate.ml:1303`](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/src/Translate.ml#L1303). 

**Priority / impact**: it silently corrupts verification status attribution downstream in [probe-aeneas](https://github.com/Beneficial-AI-Foundation/probe-aeneas).

**Where was this seen?**: reduced from [SparsePostQuantumRatchet-verify](https://github.com/Beneficial-AI-Foundation/SparsePostQuantumRatchet-verify). Concretely on `spqr::encoding::polynomial::{impl spqr::encoding::Decoder for spqr::encoding::polynomial::PolyDecoder}::decoded_message` and `spqr::kdf::hkdf_to_slice`, both marked opaque via Charon's `--opaque`.

**Regression?**: no.

---

🤖 _Co-Authored-By: Claude Opus 4.8 (1M context)_

oliver-butterley:
Previously there was an error with the emit-json feature which meant that the lean name written to translate.json would incorrectly include the crate namespace for external objects. This PR mends this problem by feeding through the relevant info about being inside a namespace (`in_namespace`) for a given file and so corrects the output. 

fixes #1248


oliver-butterley:
I wanted to add a test for this using the standard test framework within this repo. However `translate.json` includes the git commit of the version of Aeneas used to generate it and this will always be outdated when committed. 

The problem would go away if we modified the feature in either of the following ways:
- translate.json didn't include the Aeneas commit
- another command line arg which allows us to choose if translate.json includes the version info

@sonmarcho would you like either of these implemented? (Surely as a separate PR.)

sonmarcho (approved):

PR #1255, opened 2026-08-11
Add lsp info for impl_def

### LINKED ISSUE #1247: Bug:  `impl_def` does not register declaration ranges, so generated trait instances are invisible to go-to-definition and to environment-walking tools
This is a bug in the Lean backend library: the `impl_def` command elaborator in `backends/lean/`, not in the Rust → LLBC → Lean translation. It needs no `.llbc` to reproduce, and the Rust below is only to show how one obtains an `impl_def` in practice. 

This is `tests/src/defaulted_method.rs`, already in the repository, no new input needed. The relevant part is the trait with a provided method and the impl that does *not* override it:

```rust
trait Trait {
    fn provided_method(&self) -> u32 {
        self.required_method()
    }
    fn required_method(&self) -> u32;
}

struct NoOverride;
impl Trait for NoOverride {
    fn provided_method(&self) -> u32 { 73 }   // overrides everything
    fn required_method(&self) -> u32 { 12 }
}

struct YesOverride;
impl Trait for YesOverride {
    fn required_method(&self) -> u32 { 42 }   // inherits `provided_method`
}
```

**Charon command**: not applicable: the bug is in the Lean elaborator rather than in translation.

**Aeneas command**: as above.

**Versions**:
- Aeneas: `6e167c9b63a4dafd66d0e0edd9d94669f957ff7b` (`main`, tagged `nightly-2026.07.25`)
- Charon: not needed
- Lean: `v4.31.0` 

**Aeneas output**: no warnings or errors. Translation succeeds; the generated Lean
compiles. The defect is in metadata the elaborator does not record.

**Lean translation**:

The two impls in that one file are translated by two different commands, which is the whole of the bug. `NoOverride`, which overrides every method, becomes a plain `def` (`tests/lean/DefaultedMethod.lean:47`):

```lean
/-- Trait implementation: [defaulted_method::{impl defaulted_method::Trait for defaulted_method::NoOverride}]
    Source: 'tests/src/defaulted_method.rs', lines 10:0-17:1 -/
@[reducible]
def NoOverride.Insts.Defaulted_methodTrait : Trait NoOverride := {
  provided_method := NoOverride.Insts.Defaulted_methodTrait.provided_method
  required_method := NoOverride.Insts.Defaulted_methodTrait.required_method
}
```

`YesOverride`, which inherits `provided_method` and is therefore self-referential, becomes `impl_def` (`tests/lean/DefaultedMethod.lean:66`):

```lean
/-- Trait implementation: [defaulted_method::{impl defaulted_method::Trait for defaulted_method::YesOverride}]
    Source: 'tests/src/defaulted_method.rs', lines 20:0-24:1 -/
@[reducible]
impl_def YesOverride.Insts.Defaulted_methodTrait : Trait YesOverride := {
  provided_method := Trait.provided_method.default
    YesOverride.Insts.Defaulted_methodTrait
  required_method := YesOverride.Insts.Defaulted_methodTrait.required_method
}
```

The first gets an entry in Lean's `declRangeExt`; the second gets none. Note that Aeneas *has* the position: it writes `lines 20:0-24:1` into the docstring, it just never hands it to Lean.

**Explain the bug**:

`impl_def`'s elaborator never calls `addDeclarationRanges`, so no row is written to `declRangeExt`.

A declaration with no range is indistinguishable from a kernel-synthesized constant that genuinely has no source position, even though it sits at an ordinary line of a generated file.

Minimal self-contained check, no Rust or Charon involved:

```lean
import Aeneas.Tactic.Elab.TraitDefault.Init

namespace Repro

structure Trait where
  N : Nat
  P : Nat

@[trait_default]
def Trait.P.default (self : Trait) : Nat := self.N + 3

def viaDef : Trait := { N := 5, P := 3 }

impl_def viaImplDef : Trait := {
  N := 5
  P := Trait.P.default viaImplDef
}

end Repro

open Lean Meta in
#eval show MetaM Unit from do
  let env ← getEnv
  for n in [`Repro.viaDef, `Repro.viaImplDef] do
    IO.println s!"{n}"
    IO.println s!"  environment entry : {repr ((env.find? n).map ConstantKind.ofConstantInfo)}"
    IO.println s!"  declRangeExt entry: {(declRangeExt.find? env n).isSome}"
```

```
Repro.viaDef
  environment entry : some (Lean.ConstantKind.defn)
  declRangeExt entry: true
Repro.viaImplDef
  environment entry : some (Lean.ConstantKind.defn)
  declRangeExt entry: false
```

Both are ordinary `defn` constants; only the range differs.

The root cause is in [backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean (lines 319–390)](https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean#L319-L390). The elaborator destructures `declModifiers`, `declId`, `optDeclSig` and `declVal`, so it has the syntax in hand, but goes from the elaborated `Expr` straight to:

```lean
        let docCtx := (← getLCtx, ← getLocalInstances)
        addAndCompileNonRec docCtx preDef
```

`addAndCompileNonRec` takes a `PreDefinition`; it carries no syntax and cannot record positions. Lean's builtin `def` registers explicitly instead, in `elabMutualDef` (`Lean/Elab/MutualDef.lean:1260`):

```lean
for view in views, declId in expandedDeclIds do
  addDeclarationRangesForBuiltin declId.declName view.modifiers.stx view.ref
```

**Priority / impact**:

Go-to-definition does not work on any `impl_def`-declared name, since `Lean/Server/GoTo.lean` resolves through `findDeclarationRanges?`, which reads `declRangeExt`. This is the same thing PR #1011 ("Update `#decompose` to support hovering over ids and goto defs") fixed for `#decompose`; the helper it added,
`registerDecomposeInfo` in `Aeneas/Command/Decompose.lean`, is documented as "Declaration ranges (for go-to-definition on introduced names)".

Separately, tools that walk the environment drop these declarations. Using "has a declaration range" as a proxy for "corresponds to user-visible source" is reasonable, because kernel-generated constants have no range; but it silently discards every `impl_def`. In our case a generated `Iterator` instance for `Map<I, F>` has `next := sorry` (a local patch for #1043), and because the instance is declared with `impl_def` our analysis tool emitted no node for it, so the `sorry` never propagated.

**Where was this seen?**: a Rust project translated with Aeneas (SparsePostQuantumRatchet), in the `Iterator` instance for `core::iter::Map`. Reduced to `tests/src/defaulted_method.rs` from your own test suite for this report.

**Regression?**: no

**Suggested fix**:

The idiom is already in the repository, in `registerDecomposeInfo` (PR #1011). 

---

🤖 _Co-Authored-By: Claude Opus 4.8 (1M context)_

oliver-butterley:
`impl_def`-generated instances now behave like ordinary `def`-generated ones in the sense of hover over info and jump to def.

fixes #1247


backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean
```diff
@@ -389,4 +389,8 @@ elab mods:declModifiers "impl_def " id:declId sig:optDeclSig val:declVal : comma
         let docCtx := (← getLCtx, ← getLocalInstances)
         addAndCompileNonRec docCtx preDef
 
+        -- Register LSP info for the introduced name.
+        Elab.addDeclarationRangesFromSyntax declName (← getRef) id
+        Term.addTermInfo' id (← mkConstWithLevelParams declName) (isBinder := true)
+
```
mpenciak, line 394: I think all we need is the `addDeclarationRangesFromSyntax` to fix the issue/get the LSP to recognize the decl.

`addAndCompileNonRec` already handles the `addTermInfo'` (it calls `addPreDefInfo` under the hood), so the second line is redundant.

mpenciak (changes requested):
Added a comment. Otherwise looks good, and should fix the `probe-aeneas` issue!

backends/lean/Aeneas/Tactic/Elab/TraitDefault/Init.lean
oliver-butterley: OK! Really grateful for you looking at this, it's not something I thought about before. 

Here I just imitated what is already done for decompose: https://github.com/AeneasVerif/aeneas/blob/daa85d7e89400fa978be83fedbc7e475a83f0889/backends/lean/Aeneas/Command/Decompose.lean#L1421-L1422

Why in that case is the extra part required?

mpenciak: I tried tracing through to see if it's also unnecessary in the `#decompose` command... I'm not sure, because declarations are added using `addDecl` I'm not sure if the `addTermInfo'` happens in that chain of calls.

By using `addDeclarationRangesFromSyntax` it definitely does though, so you should be good deleting it in this case.

oliver-butterley: I deleted it and it seems to be fine.

oliver-butterley:
> Added a comment. Otherwise looks good, and should fix the `probe-aeneas` issue!

@mpenciak I made the changes but it still tells me that "changes requested". Is there something else to change?

mpenciak (approved):

mpenciak:
> > Added a comment. Otherwise looks good, and should fix the `probe-aeneas` issue!
> 
> @mpenciak I made the changes but it still tells me that "changes requested". Is there something else to change?

I think I just needed to approve, but it looks good now!

PR #1267, opened 2026-08-16
Generalize step to registered specifications

andricicezar:
`step` assumed the goal was `Std.WP.spec program post`, applied to exactly three arguments. This PR generalizes `step` to work with other specs. I added a test file: `Tests/Triple.lean`

Along the way:
- `analyzeTarget` now reads the goal with `getMainTarget`, which instantiates the meta-variables. A tactic such as `by_cases` leaves the goal's type as an assigned meta-variable, and `step*` would stop at such a goal instead of walking into the branch below it.
- `prettyMonadEq` is generalized over the monad, so that `let*` keeps recording where its outputs come from outside of `Std.Result`.
- the tactic given by the user after `by` is also used to close the final goal.
- the decomposition of a specification statement is factored out into `getSpecInfoArgs`, `getSpecProgram` and `getSpecPost`, which had been duplicated at five call sites.

The PR was AI assisted, but I scrutinized thoroughly. However, I am new to Aeneas.


sonmarcho:
Adding a correction: `step` does not assume the goal is `Std.WP.spec program post` (it is supposed to be general) but there are indeed places that we forgot to update.

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -544,9 +555,11 @@ def tryMatch (info : SpecInfo) (lifting : Option LiftingInfo) (isLet : Bool) (th
   let specMonoBindTy ← inferType specMonoBind
   trace[Step] "Applied specMonoBind with theorem: {specMonoBind}: {specMonoBindTy}"
 
-  let (specMonoBindMVars, _, specMonoBindTy) ← forallMetaBoundedTelescope specMonoBindTy 1
-  if (specMonoBindMVars.size ≠ 1) then throwError "Unreachable"
-  let ngoal := specMonoBindMVars[0]!.mvarId!
+  let (specMonoBindMVars, _, specMonoBindTy) ←
+    forallMetaBoundedTelescope specMonoBindTy (countLeadingForalls specMonoBindTy)
```
sonmarcho: Going from 1 forall to any number of forall is suspicious (and I think if we do so we don't need `forallMetaBoundedTelescope` but can directly use `forallMetaTelescope`)

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -565,7 +578,8 @@ def tryMatch (info : SpecInfo) (lifting : Option LiftingInfo) (isLet : Bool) (th
   mgoal.assign specMonoBind
   trace[Step] "New goal: {ngoal}"
 
-  let mvarsIds := mvars.map Expr.mvarId!
+  let extraPreconditions := specMonoBindMVars.pop.map Expr.mvarId!
```
sonmarcho: We should move this just after ngoal (otherwise it's hard to understand why we `pop`)

backends/lean/Aeneas/Tactic/Step/Step.lean
```diff
@@ -589,8 +603,10 @@ def introPrettyEquality (args : Args) (fExpr : Expr) (outputFVars : Array Expr)
   -- Construct the tuple of outputs
   let pat ← mkProdsVal outputFVars.toList
   trace[Step] "Constructed the pattern: {pat}"
-  -- Create the equality
-  let e ← mkAppM ``eq_imp_prettyMonadEq #[fExpr, pat]
+  -- Create the equality.
+  let some e ← observing? (mkAppM ``eq_imp_prettyMonadEq #[fExpr, pat])
```
sonmarcho: No need for `observing?`. Also pay attention to the noise when updating comments (by adding a `.` for instance).

backends/lean/Aeneas/Tactic/Step/StepStar.lean
```diff
@@ -385,25 +379,21 @@ inductive TargetKind where
...
-      else
-        pure .result
+    let goalTy ← getMainTarget
+    -- Dive into a registered specification statement.
+    let some program ← observing? (Step.getSpecProgram goalTy)
+      | trace[Step] "not an application of a registered specification statement: {goalTy}"
+        return .result
+    trace[Step] "application of a registered specification statement about: {program}"
+    let e ← Utils.normalizeLetBindings program
+    if let .const ``Bind.bind .. := e.getAppFn then
```
sonmarcho, line 390: The comment `-- Check if this is a bind` got removed (we should pay attention to preserve comments)

backends/lean/Aeneas/Tactic/Step/StepStar.lean
```diff
@@ -385,25 +379,21 @@ inductive TargetKind where
...
-        let #[_m, _self, _α, _β, _value, cont] := e.getAppArgs
-          | throwError "Expected bind to have 6 arguments, found {← e.getAppArgs.mapM (liftM ∘ ppExpr)}"
-        let names ← Step.getPostNames cont
-        pure (.bind names)
-      else if let .some bfInfo ← Bifurcation.Info.ofExpr e then
-        pure (.switch bfInfo)
-      else
-        pure .result
+    let goalTy ← getMainTarget
+    -- Dive into a registered specification statement.
```
sonmarcho: Update the comment

backends/lean/Aeneas/Tactic/Step/StepStar.lean
```diff
@@ -385,25 +379,21 @@ inductive TargetKind where
...
+    let e ← Utils.normalizeLetBindings program
+    if let .const ``Bind.bind .. := e.getAppFn then
+      let #[_m, _self, _α, _β, _value, cont] := e.getAppArgs
+        | throwError "Expected bind to have 6 arguments, found {← e.getAppArgs.mapM (liftM ∘ ppExpr)}"
+      let names ← Step.getPostNames cont
+      pure (.bind names)
+    else if let .some bfInfo ← Bifurcation.Info.ofExpr e then
+      pure (.switch bfInfo)
     else
-      trace[Step] "not an application of `spec` with arity 3"
```
sonmarcho, line 406: Keep the trace

sonmarcho (commented):
Thanks! I opened a few comments: I'm closing them myself to merge the PR asap (I'll let you look at them to see the modifications I made - a few nits and technical details).

PR #1268, opened 2026-08-17
doc: add docstring to #decompose command

mpenciak:
Adds a docstring for the `#decompose` command so that the LSP hover text gives hints at usage. Includes a link to the module docstring which has much more detail.


backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -281,6 +281,44 @@ partial def elabDecomposePat : Syntax → Except String DecomposePattern
...
+- `afterLets (pat)`: navigate past all leading let/bind bindings to the terminal expression,
+  then apply `pat` (`letAt N` without having to know `N`).
+- `branch i (pat)`: navigate into branch `i` of an `if-then-else`, `dite` or `match`
+  (0 is `then` / the first alternative), then apply `pat`. Match-pattern lambdas are opened
+  automatically.
+- `lam n (pat)`: open `n` lambda binders, then apply `pat`.
+- `appFun (pat)` / `argArg i (pat)`: navigate into the function, resp. argument `i`, of an
+  application.
+
+Full documentation can be found in the module docstring at
```
oliver-butterley: This might be rather subjective... 🤣  I don't love the weblink here, I'm very keen to keep everything local. How about we just drop this? I checked and you can "go to def" on #decompose and that takes you to the file which has the full info in the module docstrings. The hover text give sufficient info so that users can get an idea and be reminded of the syntax, ones who want more detail can be expected to go to the file where it is defined?

On the other hand, what do you think about adding a note in `backends/lean/Aeneas/Command/Decompose.lean` pointing the user to all the test files which contain great usage examples?

oliver-butterley (approved):
Really useful to users to have this info easily viewable!

backends/lean/Aeneas/Command/Decompose.lean
sonmarcho: I agree: let's just mention that the full documentation is available in the module that defines `#decompose`

backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -281,6 +281,44 @@ partial def elabDecomposePat : Syntax → Except String DecomposePattern
...
+
+Matches on the patterns `pattern₁`, `pattern₂`, ..., introduces the definitions `auxName₁`,
+`auxName₂`, ... and provides the theorem `eqThm` grouping them all together into `originalFn`.
+
+## Available Patterns
+
+A pattern navigates to a position in the body; the extraction happens there.
+
+- `letRange start count`: extract `count` consecutive let/bind bindings starting at index
+  `start` (0-indexed).
```
sonmarcho:
```suggestion
  `start` (starting at 0).
```

backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -281,6 +281,44 @@ partial def elabDecomposePat : Syntax → Except String DecomposePattern
...
+- `full`: extract the whole expression at the current position.
+- `letAt i (pat)`: navigate to the value of binding `i`, then apply `pat`. When `i` points
+  past all bindings, `pat` applies to the terminal expression instead.
```
sonmarcho:
```suggestion
- `letAt i (pat)`: navigate to the value of binding `i`, then apply `pat`.
```

backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -281,6 +281,44 @@ partial def elabDecomposePat : Syntax → Except String DecomposePattern
...
+- `afterLets (pat)`: navigate past all leading let/bind bindings to the terminal expression,
+  then apply `pat` (`letAt N` without having to know `N`).
```
sonmarcho:
```suggestion
- `afterLets (pat)`: navigate past all leading let/bind bindings to the terminal expression,
  then apply `pat`.
```

backends/lean/Aeneas/Command/Decompose.lean
mpenciak: Sounds good! I took out the link and added a note to point people to the tests.

oliver-butterley (approved):
LGTM!

sonmarcho (approved):

PR #1269, opened 2026-08-18
Go-to-Rust-Def

mpenciak:
This PR adds a hook into the Lean LSP implementation that points an extracted Lean declaration to its associated Rust definition.

The implementation adds a `@[rust_source ...]` attribute that points to the file, line, and column for the rust source declaration. This adds the file information into an environment extension. This extension is queried when the language server receives a go-to-def request, and serves the Rust file as an additional option in the Go-To-Def selection.

The feature is gated behind a `-rust-source-links` Aeneas flag. 

I've also investigated whether it's possible to offer the rust definition as an option when you try to go-to-def on the Lean declaration itself (not just its usage sites). It seems possible, but it would be a more intrusive hook into the LSP so I'm not sure it's worth going down that path.


backends/lean/Aeneas/Extract/Extract.lean
```diff
@@ -1,4 +1,6 @@
 import AeneasMeta.Extensions
+/- We need to import this here so it's available in extracted files with `import Aeneas` -/
+import AeneasMeta.RustSource.GoToDef
```
sonmarcho: I think we should just add `import AeneasMeta` in the top-level `Aeneas.lean`

backends/lean/AeneasMeta.lean
```diff
@@ -2,6 +2,8 @@ import AeneasMeta.Async
 import AeneasMeta.BvEnumToBitVec
 import AeneasMeta.Extensions
 import AeneasMeta.OptionConfig
+import AeneasMeta.RustSource
+import AeneasMeta.RustSource.GoToDef
```
sonmarcho: I suggest importing `GoToDef` from `RustSource.lean` (that's the way the Lean project is structured) so that here we only need to import `RustSource`. I think it involves reorganizing a bit the files in the present case (GoToDef imports RustSource).

src/Main.ml
```diff
@@ -139,6 +139,10 @@ let () =
       ( "-emit-json",
         Arg.Set emit_json,
         " Emit a translation.json file alongside the Lean files (Lean only)" );
+      ( "-rust-source-links",
```
sonmarcho:
```suggestion
      ( "-rust-source-spans",
```

src/Main.ml
```diff
@@ -139,6 +139,10 @@ let () =
...
+        Arg.Set rust_source_links,
```
sonmarcho: I prefer "spans" to "links"

tests/lean/RustSource.lean
```diff
@@ -0,0 +1,124 @@
...
+set_option maxHeartbeats 1000000
+
+/- You can set the `maxRecDepth` value with the `-max-recdepth` CLI option -/
+set_option maxRecDepth 2048
+
+namespace rust_source
+
+/-- [rust_source::Pair]
+    Visibility: public -/
+@[rust_source "tests/src/rust_source.rs" 9 0 12 1]
```
sonmarcho: Would it be possible to have a nicer syntax, to see "line", "column", etc. as when it is printed in the docstring?

sonmarcho (commented):
Nice, thanks! Only minor comments

mpenciak:
I think I addressed all your comments!

PR #1272, opened 2026-08-19
feat: reduce Errors to only panic and undef

abentkamp:
This PR reduces the Error type to just two cases: undef and panic.

I used Claude to do this, but a simple search and replace would have had the same effect. I reviewed the changes and checked that the entire codebase no longer contains the strings
`assertionFailure`, `integerOverflow`, `divisionByZero`, `arrayOutOfBounds`, `maximumSizeExceeded`.

Moreover, I replaced integer overflows with `undef` instead of `panic` whenever the functions behavior depends on whether it is compiled in debug or release mode.

### DIFF
--- a/backends/lean/Aeneas/Std/Primitives.lean
+++ b/backends/lean/Aeneas/Std/Primitives.lean
@@ -49,11 +49,6 @@ def elabImpl : CommandElab := fun (stx: Syntax) => do
 -/
 
 inductive Error where
-   | assertionFailure: Error
-   | integerOverflow: Error
-   | divisionByZero: Error
-   | arrayOutOfBounds: Error
-   | maximumSizeExceeded: Error
    | panic: Error
    | undef: Error
 deriving Repr, BEq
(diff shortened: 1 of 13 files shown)


ia0 (approved):
This is supposed to address #1225 right?

If yes, then we probably want some of the `integerOverflow` to be `undef` rather than `panic` according to what @sonmarcho [suggested](https://github.com/AeneasVerif/aeneas/issues/1225#issuecomment-5071096581).

abentkamp:
Ah, thanks, I didn't know about that discussion!

abentkamp:
By the same logic `debug_assert`, which gets translated as `massert` would have to be modeled as undefined behavior, right?

ia0:
> By the same logic `debug_assert`, which gets translated as `massert` would have to be modeled as undefined behavior, right?

Interesting, I didn't think of that but it makes sense.

backends/lean/Aeneas/Std/Primitives.lean
```diff
@@ -49,11 +49,6 @@ def elabImpl : CommandElab := fun (stx: Syntax) => do
 -/
 
 inductive Error where
-   | assertionFailure: Error
-   | integerOverflow: Error
-   | divisionByZero: Error
-   | arrayOutOfBounds: Error
-   | maximumSizeExceeded: Error
    | panic: Error
    | undef: Error
```
sonmarcho: Should we simply rename this to `ub`?

abentkamp: Done.

abentkamp:
I don't understand the CI error. When I run `make extract-lean-std`, the files remain unchanged.

abentkamp:
I rebased on main. That should hopefully fix CI.

PR #1284, opened 2026-08-23
Fix name clash when a struct constructor is passed as a function

### LINKED ISSUE #767: Bug: name clash bug when struct constructor is passed as a function pointer
When a tuple struct constructor is passed as a function pointer using `.map()` Aeneas generates a standalone constructor function whose name clashes with the type definition. Both get the same name in the generated Lean code.

## MWE:

```rust
pub struct Wrapper(pub [u8; 32]);

/// This version FAILS: passing the constructor as a function pointer causes Aeneas to generate a 
/// standalone constructor function that clashes with the type name.
pub fn make_wrapper_broken(x: Result<[u8; 32], ()>) -> Result<Wrapper, ()> {
    x.map(Wrapper)
}

/// This version WORKS: using a closure avoids generating the standalone constructor function, so 
/// no name clash occurs.
pub fn make_wrapper_fixed(x: Result<[u8; 32], ()>) -> Result<Wrapper, ()> {
    x.map(|b| Wrapper(b))
}
```

## Aeneas error message
```text
[Error] Name clash detected: the following identifiers are bound to the same
name "Wrapper" (the generated code will be incorrect):
- type name: name_clash_mwe::Wrapper
- name_clash_mwe::Wrapper

[Error] Error when registering the name for id: name_clash_mwe::Wrapper:
The chosen name is already in the names set: Wrapper
```

**Aeneas version**: commit `74749a0`

## Real-world occurrence

This pattern appears in the curve25519-dalek crate:
- [curve25519-dalek/src/edwards.rs](https://github.com/dalek-cryptography/curve25519-dalek/blob/61533d75cfc5df3452cbee20181c89ab2e40bccd/curve25519-dalek/src/edwards.rs#L424)
- [curve25519-dalek/src/ristretto.rs](https://github.com/dalek-cryptography/curve25519-dalek/blob/61533d75cfc5df3452cbee20181c89ab2e40bccd/curve25519-dalek/src/ristretto.rs#L256)

## Workaround

As per the "fixed" function above.

Boarders:
When a struct constructor name is used as a first-class value (e.g. `x.map(Newtype)`), Charon emits a standalone constructor function whose name coincides with the type's name. This causes a name clash error in the Lean backend as type names are not sanitized with a "_t" suffix.

This fixes the issue by detecting it during extraction and instead emitting `<Type>.constructor`

Add tests/src/name_clash_767.rs.

Closes #767.

AI disclosure: I used AI to ask questions about the codebase and for checking various output as I went, but not for writing any of the code.

Note on design: In the generated lean it now generates a constructor function which is named after the type if that type happens to be used as a standalone function as in the test, so, for example:

```rust
pub struct Wrapper(pub [u8; 32]);
```

will generate, if the constructor is used as a function, the lean code:

```lean
def Wrapper.constructor (a : Array Std.U8 32#usize) : Result Wrapper := do
  ok a
```

In the code I use `<Typename>.constructor` but pessimistically call `basename_to_unique` on "constructor" in case the user happens to have some method called `constructor`. If this is the wrong way to handle that kind of issue with an internally generated name, then I'm more than happy to change it.


tests/src/name_clash_767.rs
```diff
@@ -0,0 +1,17 @@
...
+
+// Broken: `Wrapper` is passed as a first-class function pointer.
+pub fn make_wrapper_broken(x: Result<[u8; 32], ()>) -> Result<Wrapper, ()> {
+    x.map(Wrapper)
+}
+
+// Workaround: wrapping in a closure avoids synthesizing the constructor fn.
+pub fn make_wrapper_fixed(x: Result<[u8; 32], ()>) -> Result<Wrapper, ()> {
+    x.map(|b| Wrapper(b))
+}
```
oliver-butterley: Is there a benefit to including this function which previously was already fine? 

Seems best to just include `make_wrapper_broken` which was a problem and now isn't.

Also, the suffix `_broken` could be removed.

Boarders: good point, I'll remove it

tests/src/name_clash_767.rs
```diff
@@ -0,0 +1,11 @@
+//@ [!lean] skip
+//! Reproducer for https://github.com/AeneasVerif/aeneas/issues/767
+//!
+//! Using a tuple-struct constructor as a function (`x.map(Wrapper)`) makes
+//! Aeneas synthesize a standalone constructor function whose generated name
+//! clashes with the type definition `Wrapper`.
```
oliver-butterley:
```suggestion
//! Using a tuple-struct constructor as a function (`x.map(Wrapper)`) could,
//! if Aeneas uses a naive naming scheme, be assigned a name
//! which clashes with the type definition `Wrapper`.
```

Or some other variant of the comment which won't become stale.

src/extract/ExtractBase.ml
```diff
@@ -2415,9 +2427,24 @@ let ctx_compute_fun_global_name_no_suffix (item_meta : T.item_meta)
...
+  | Lean -> (
+      match fun_decl_constructor_target item_meta ctx with
+      | Some tdecl ->
+          (* emit new name for Struct constructor, see `fun_decl_constructor_target` *)
+          let tname =
+            ctx_compute_type_name tdecl.item_meta ctx tdecl.item_meta.name
+          in
+          let cons_name = tname ^ ".constructor" in
+          basename_to_unique ctx cons_name
+      | _ ->
```
oliver-butterley:
```suggestion
      | None ->
```

src/extract/ExtractBase.ml
```diff
@@ -2415,9 +2427,24 @@ let ctx_compute_fun_global_name_no_suffix (item_meta : T.item_meta)
...
-    ~is_trait_decl_field ctx
+  match backend () with
+  | Lean -> (
+      match fun_decl_constructor_target item_meta ctx with
+      | Some tdecl ->
+          (* emit new name for Struct constructor, see `fun_decl_constructor_target` *)
+          let tname =
+            ctx_compute_type_name tdecl.item_meta ctx tdecl.item_meta.name
+          in
+          let cons_name = tname ^ ".constructor" in
```
oliver-butterley: Would it make sense to immitate the code style of `mk_lean_struct_constructor` to define this addition of a suffix rather than inline it?

src/extract/ExtractBase.ml
```diff
@@ -2415,9 +2427,24 @@ let ctx_compute_fun_global_name_no_suffix (item_meta : T.item_meta)
...
+          basename_to_unique ctx cons_name
```
oliver-butterley: It seems very likely that basename_to_unique is required but I can't think of the case where this happens. Do you have a case in mind?

src/extract/ExtractBase.ml
```diff
@@ -1465,6 +1465,18 @@ let fun_decl_kind_to_post_qualif (kind : decl_kind) : string option =
...
+
+    A struct constructor used as a first-class value (e.g. [x.map(Point)]) is
+    emitted by Charon as a standalone function whose name coincides with the
+    type's. We use this at extraction time to give it a non-clashing name. *)
+let fun_decl_constructor_target (item_meta : T.item_meta) (ctx : extraction_ctx)
+    : Types.type_decl option =
+  let check_item (ty_decl : Types.type_decl) : bool =
+    ty_decl.item_meta.is_local && item_meta.name = ty_decl.item_meta.name
+  in
+  ctx.crate.type_decls |> TypeDeclId.Map.values |> List.find_opt check_item
```
oliver-butterley: If I understand all of this correctly (which is far from certain!) this line is of O(#types) and runs for each function so together this functionality is O(#funcs × #types). It seems that it should be possible to do the naming in a more efficient way. (Although I really don't know how significant this is for large crates.)

Would it be possible to imitate what is done by `types_filter_type_args_map` in `extraction_ctx` and therefore precompute what will be required here (let's call it `local_type_decls_by_name` for this discussion). This is then computed once:
```
let local_type_decls_by_name =
  TypeDeclId.Map.values crate.type_decls
  |> List.filter (fun (td : Types.type_decl) -> td.item_meta.is_local)
  |> List.fold_left (fun m td -> NameMap.add td.item_meta.name td m) NameMap.empty

```
and then at this point in the code we only need to use the look up `NameMap.find_opt item_meta.name ctx.local_type_decls_by_name`.

However, I don't have the mastery of this code that @sonmarcho has so it is possible that I'm missing things!

oliver-butterley (commented):
Looks great! Thanks for doing this. The test is perfect and the functionality achieved is as it should be. 

I made a few comments which came to mind, If they aren't useful then ignore them. 

In terms of precisely where in the code the naming logic should live, I realise I don't know the subtle details of this code sufficiently well and we'll need to get an opinion from @sonmarcho.

src/extract/ExtractBase.ml
Boarders: this is intended to guard against a case like this:
```rust
struct Wrapper(u32);

impl Wrapper {
    fn constructor(x: u32) -> u32 {
        x + 1
    }
}

fn make(x: Option<u32>) -> Option<Wrapper> {
    x.map(Wrapper)
}
```

src/extract/ExtractBase.ml
Boarders: have now made that change, makes sense to me for the (perhaps slight) efficiency - thanks!

src/extract/ExtractBase.ml
```diff
@@ -2427,9 +2433,36 @@ let ctx_compute_fun_global_name_no_suffix (item_meta : T.item_meta)
...
 let ctx_compute_fun_name_no_suffix (item_meta : T.item_meta) (src : fun_source)
     ~(is_trait_decl_field : bool) (ctx : extraction_ctx) : string =
-  ctx_compute_fun_global_name_no_suffix item_meta
-    (extractable_fun_source ctx src)
-    ~is_trait_decl_field ctx
+  (* A struct constructor used as a first-class value (e.g. [x.map(Point)]) is
+    emitted by Charon as a standalone function whose name coincides with the
+     type's name in lean as that backend doesn't add a type suffix
+     (see: ctx_compute_type_name). *)
+  let local_type_decls_by_name =
```
mpenciak: Is it possible to construct this outside of `ctx_compute_fun_name_no_suffix`? This function gets called a bunch, so we'd be reconstructing the NameMap on each call. Could we have a prepass and have it contained in the `extraction_ctx`?

src/extract/ExtractBase.ml
```diff
@@ -2427,9 +2433,36 @@ let ctx_compute_fun_global_name_no_suffix (item_meta : T.item_meta)
...
+    TypeDeclId.Map.values ctx.crate.type_decls
+    |> List.filter (fun (td : Types.type_decl) -> td.item_meta.is_local)
+    |> List.fold_left
+         (fun m (td : Types.type_decl) -> NameMap.add td.item_meta.name td m)
+         NameMap.empty
+  in
+  let mk_lean_standalone_constructor_fn name = name ^ ".constructor" in
+  match backend () with
+  | Lean -> (
+      match NameMap.find_opt item_meta.name local_type_decls_by_name with
```
mpenciak: I'm worried that matching on the `Name` alone might cause some mistranslation. For example, I believe this is valid Rust:

```rust
pub struct Foo {}
pub fn Foo() -> u32 { 0 }
```
but the extraction of the `Foo()` function would be `Foo.constructor` based purely on the name collision. 

I don't have a handle the extraction logic well enough to suggest a fix, but is it possible to key on whether it's a constructor and have this logic fire on name conflicts only then instead?

mpenciak (commented):
I added a few comments.

One additional note is that the test should be added to the tests lakefile, and the "naming convention" we've been using so far would be something like `issue-767-name-clash.rs`.

Also the CI failure can be fixed with a rebase I believe, so after you address the comments I believe it should be good to go!

src/extract/ExtractBase.ml
Boarders: I checked, and this is valid rust and was giving me incorrect code gen on this branch. So now I have made it check whether the identifier is a struct constructor when registering the name. That meant that I added a test like this:

```rust
//@ [!lean] skip
//@ [lean] known-failure
pub struct Foo {
    pub x: u32,
}

pub fn Foo() -> u32 {
    0
}

pub fn use_foo() -> u32 {
    Foo()
}
```

which currently on this branch produces:
```lean
def Foo := Unit
def Foo.constructor : Result Std.U32 := do   -- ← the plain fn Foo, mislabeled
  ok 0#u32
def use_foo : Result Std.U32 := do
  Foo.constructor
```

After fixing it, it now is back to giving a name clash:

```bash
- type name: issue_767_name_clash_fn::Foo
  Source: 'tests/src/issue-767-name-clash-fn.rs', lines 12:0-14:1
- issue_767_name_clash_fn::Foo
  Source: 'tests/src/issue-767-name-clash-fn.rs', lines 16:0-18:1
  ```

This is reasonable and would require more careful thinking. If someone were to write the diabolical pattern you bring up, then they can fix it by using [#[aeneas::rename("NAME")] as the hint to that error suggests. That was not possible with a struct constructor since there is nowhere to add such an annotation, so I think this is still a reasonable place for this PR to end up.

src/extract/ExtractBase.ml
Boarders: Re-factored this so the map is computed only once

src/extract/ExtractBase.ml
Boarders: This is added as a known failing test and I can add a follow-up issue if you think it is worth it (though as I said, this is much easier for a user to fix that 767)

sonmarcho:
Thanks for this! I noticed it is actually possible to update Charon to provide the information that the function was generated from an ADT application, leading to a simpler PR, so I'm closing this one:
https://github.com/AeneasVerif/charon/pull/1407
https://github.com/AeneasVerif/aeneas/pull/1330

PR #1286, opened 2026-08-24
fix: fix default impl of Ord::max,min

abentkamp:
The implementations of `Ord::max` and `Ord::min` are wrong for cases where two values are equal w.r.t. the order, but still distinguishable.

Here is a playground exhibiting such a case:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=395a33a2f9fdcfd20a414ac9ebbdce65

Here are the corresponding lines in the Rust implementation of  `Ord::max` and `Ord::min`.
https://doc.rust-lang.org/src/core/cmp.rs.html#1002-1078

The changes to `core.cmp.impls.Ord'S.min` and `core.cmp.impls.Ord'S.max` are needed to fix the proofs.


sonmarcho:
For some reason I can't merge main into this branch? @abentkamp could you do it?

abentkamp:
> For some reason I can't merge main into this branch? @abentkamp could you do it?

Done.

sonmarcho:
I think you need to regenerate `ExtractBuiltinLean.ml` with `gmake extract-lean-std`

abentkamp:
Just merging `main` did the trick.

sonmarcho:
Can you merge main again? ^^'
It seems another PR took the priority...

Issue #706, opened 2026-01-16
Reorganize the files in the Lean backend


sonmarcho:

oliver-butterley:
Among other things, perhaps:
- Use a separate `lean_lib` called `AeneasTest` as a home for the hand written Lean tests.
- Include the lean files created from Rust in another `lean_lib`. 
- Use a single lakefile at the root of the repo
- Several core files could be split into a `Basic` for the defs and `Lemmas` / etc. for the various related theorems. This would reduce the size of files and also separate the Lean defs which are models of Rust from all the lean results which are derived from those defs.
- Arrange so that lake exe mk_all updates each aggregator file for all lean files in the entire repo

PR #1299, opened 2026-08-26
Add ci check for  `tests/lean/lakefile.lean`

oliver-butterley:
The file `tests/lean/lakefile.lean` needs to be updated each time a new lean test is added. This PR adds a check in CI to confirm that the lakefile is properly updated. (Also updates the lakefile because one entry was missing.) 

The part of the CI which checks that the lean files build correctly always regenerates the lakefile before building and so nothing was getting missed, this was only an issue for building the lean test files locally. 

A more general solution which fits with standard lean practice would be to have a single lakefile in root with several lean_lib, one of which is for the test output lean files (#706). Then in CI we would confirm that `lake exe mk_all` is up to date. 

AI: used to get info on the args used for diff, comments, etc written manually.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -193,7 +193,7 @@
             make format
             rm -rf ./src/_build
             rm -rf ./tests/test_runner/_build
-            if ! diff --no-dereference -ru . ${src}; then
+            if ! diff --no-dereference -ru ${src} .; then
               echo 'ERROR: Code is not formatted. Run `make format` to format the project files.'
               exit 1
             fi
--- a/flake.nix
+++ b/flake.nix
@@ -189,11 +189,18 @@
             inputs.charon.packages.${system}.rustToolchain
           ];
           buildPhase = ''
+            # Check that the OCaml code is formatted
             make format
             rm -rf ./src/_build
             rm -rf ./tests/test_runner/_build
             if ! diff --no-dereference -ru . ${src}; then
               echo 'ERROR: Code is not formatted. Run `make format` to format the project files.'
+              exit 1
+            fi
+            # Check that `tests/lean/lakefile.lean` is up to date.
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -189,13 +189,20 @@
...
             make format
             rm -rf ./src/_build
             rm -rf ./tests/test_runner/_build
             if ! diff --no-dereference -ru . ${src}; then
               echo 'ERROR: Code is not formatted. Run `make format` to format the project files.'
               exit 1
             fi
+            # Check that `tests/lean/lakefile.lean` is up to date.
+            make -C tests/lean lakefile.lean
+            if ! diff --no-dereference -u ${src}/tests/lean/lakefile.lean tests/lean/lakefile.lean; then
```
Nadrieril, line 202: This is already diffed in the diff above, maybe just move the `make` command?

oliver-butterley: Sure, we can do that. I did it this way so the message to the user explains clearly which of the two problems had happened. 

Shall I combine them in one diff and update the message to mention both problems?

flake.nix
```diff
@@ -189,13 +189,20 @@
             inputs.charon.packages.${system}.rustToolchain
           ];
           buildPhase = ''
+            # Check that the OCaml code is formatted
             make format
             rm -rf ./src/_build
             rm -rf ./tests/test_runner/_build
             if ! diff --no-dereference -ru . ${src}; then
```
Nadrieril: The order has been bothering me for a while, let's be consistent and use the same order in both 
```suggestion
            if ! diff --no-dereference -ru ${src} .; then
```

flake.nix
Nadrieril: Oh sorry, I got confused, it's clearer the way you did it, no need to change anything

PR #1301, opened 2026-08-26
Fix Lean keyword escaping

### LINKED ISSUE #1285: Bug: Lean keywords do not get escaped in binder position
Lean keywords such as `end` do not get escaped in binder position, e.g., when used as the function argument.

```rust
  pub fn f(end: u32) -> u32 {
      end
  }
```

**Charon command**: 
`charon rustc --preset=aeneas --dest-file minimal.llbc -- minimal.rs --crate-type=rlib`

**Aeneas command**: 
`aeneas minimal.llbc -backend lean -dest out`

**Versions**:
- Aeneas: 5d08da45
- Charon: 0.1.235

**Aeneas output**: no warnings or errors

**Lean translation**:
```
  def f (end : Std.U32) : Result Std.U32 := do
    ok «end»
```
The use site is escaped, the binder is not. Lean rejects it:
```
  error: unexpected token 'end'; expected '_' or identifier
  error: Invalid `end`: There is no current scope to end
```

**Regression**: This was introduced by #1169

abentkamp:
Fixes https://github.com/AeneasVerif/aeneas/issues/1285

I rewrote the escaping mechanism to happen only shortly before printing. The name maps that are used to prevent name clashes do now contain the names in unescaped form. The functions that retrieve disambiguated names from the name maps (`ctx_get` and friends) are now responsible for the escaping. For the rare case where we want to retrieve the name unescaped, I added a function `ctx_get_raw`.

While working on this, I discovered two other issues:
- The change from `(ctx, field_name)` to `(ctx, vname)` in `ExtractTypes.ml` is necessary because we want to print the names of the local variables here, not the field names. See the new `tests/src/lean-keywords-clash.rs` for an example where this matters.
- The test lakefile was missing a target for `Issue1250AnonymousConst`. The `tests/lean/script.sh` script added it automatically.

**AI disclaimer:** The code was generated by AI, I accompanied the process and reviewed it carefully.


oliver-butterley:
The French quote escaping logic is basically the same as before but moved later in the process so that is is also used for everything which comes from `/src/extract/ExtractBuiltinLean.ml` (the definitions for the standard library). Did I understand correctly? 

Since `ExtractBuiltinLean.ml` is auto generated, this means that `tests/lean/Range.lean` and/or `backends/lean/Aeneas/Extract/Extract.lean` need updating (to remove the quotes) and then `backends/lean/AeneasExtract.lean` run again (`lake exe extract`) to recreate `ExtractBuiltinLean.ml` without French quotes. At the moment the edits to `ExtractBuiltinLean.ml`  are done manually.

sonmarcho:
I just bumped into the exact same issue and was about to open a PR. This one is more complete: let me take it from here.

oliver-butterley:
As a secondary issue, I'll prepare a PR to make the CI check that `ExtractBuiltinLean.ml` and `LeanKeywords.ml` are always correctly updated.

sonmarcho:
> As a secondary issue, I'll prepare a PR to make the CI check that `ExtractBuiltinLean.ml` and `LeanKeywords.ml` are always correctly updated.

This would be super useful: I've been wanting that for a while. Thanks!

abentkamp:
Oh, I forgot that these are auto-generated, thanks!

> This one is more complete: let me take it from here.

So you'll take care of this issue with the autogenerated files @sonmarcho ?

sonmarcho:
Yes, I'm doing the changes right now

oliver-butterley:
> > As a secondary issue, I'll prepare a PR to make the CI check that `ExtractBuiltinLean.ml` and `LeanKeywords.ml` are always correctly updated.
> 
> This would be super useful: I've been wanting that for a while. Thanks!

Done in #1305

sonmarcho:
Closing in favour of https://github.com/AeneasVerif/aeneas/pull/1306

Issue #1288, opened 2026-08-24
Bug: The Lean model of `alloc::vec::insert` is wrong


abentkamp:
According to the [Rust docs](https://doc.rust-lang.org/alloc/vec/struct.Vec.html#method.insert), `alloc::vec::insert` "inserts an element at position index within the vector, shifting all elements after it to the right." The Lean model, however, overwrites the element at the given position and does not shift the elements after it.

Here is the current implementation:
```lean
@[rust_fun "alloc::vec::{alloc::vec::Vec<@T>}::insert" (keepParams := [true, false])]
def Vec.insert {α : Type u} (v: Vec α) (i: Usize) (x: α) : Result (Vec α) :=
  if i.val < v.length then
    ok ⟨ v.val.set i x, by have := v.property; simp [*] ⟩
  else
    fail arrayOutOfBounds
```

RamonJales:
can i get this issue??

abentkamp:
I think this is already fixed in #1302

PR #1302, opened 2026-08-26
`Vec::insert` translated to `.set` instead of `.insertIdx`

Raina3266:
Hello

I was recently using Aeneas for a project but noticed that `Vec::insert` seems to be translated incorrectly. I am a beginner with Aeneas and Lean, so I'm not certain, but it seems clear enough to me.

Here is the code that gets generated for `Vec::insert`: 

```
def Vec.insert {α : Type u} (v: Vec α) (i: Usize) (x: α) : Result (Vec α) :=
  if i.val < v.length then
    ok ⟨ v.val.set i x, by have := v.property; simp [*] ⟩
  else
    fail arrayOutOfBounds
```

It looks like this just called `.set` on the list, which doesn't change the length, just sets that particular index. But `Vec::insert` inserts a value at a particular index, so always either panics (if the index is out of bounds) or increases the length of the `vec` by 1.

Of course, I'm a beginner though, so happy to be corrected on any of this.

AI usage:
- I discovered the bug while working on a project with AI
- I initially wrote the PR myself, but couldn't write the proof without grind. I noticed none of the other proofs in this file use grind, so I asked an AI to rewrite it without grind

Thanks 😁


ia0:
This sounds related to #1288.

backends/lean/Aeneas/Std/Vec.lean
```diff
@@ -135,14 +135,19 @@ theorem Vec.push_spec {α : Type u} (v : Vec α) (x : α) (h : v.val.length < Us
 @[rust_fun "alloc::vec::{alloc::vec::Vec<@T>}::insert" (keepParams := [true, false])]
 def Vec.insert {α : Type u} (v: Vec α) (i: Usize) (x: α) : Result (Vec α) :=
   if i.val < v.length then
-    ok ⟨ v.val.set i x, by have := v.property; simp [*] ⟩
+    if h : v.length < Usize.max then
+      ok ⟨ v.val.insertIdx i x,
+        List.length_insertIdx_le_succ.trans (Nat.succ_le_of_lt h) ⟩
```
sonmarcho:
```suggestion
      ok ⟨ v.val.insertIdx i x,
           List.length_insertIdx_le_succ.trans (Nat.succ_le_of_lt h) ⟩
```

sonmarcho (approved):
Thanks for this!
It indeed fixes https://github.com/AeneasVerif/aeneas/issues/1288

PR #1306, opened 2026-08-26
Fix Lean keyword escaping - #1301

### LINKED ISSUE #1285: Bug: Lean keywords do not get escaped in binder position

sonmarcho:
Same PR as https://github.com/AeneasVerif/aeneas/pull/1301, that for some reason I can't push into - probably because I'm missing some access rights.

Fixes https://github.com/AeneasVerif/aeneas/issues/1285


tests/lean/LeanKeywordsClash.lean
```diff
@@ -0,0 +1,47 @@
...
+    Visibility: public -/
+inductive Rec where
+| mk : Option Rec → Std.U32 → Rec
+
+def Rec.«end» (x : Rec) := match x with | Rec.mk x1 _ => x1
+
+def Rec.v (x : Rec) := match x with | Rec.mk _ x1 => x1
+
+@[simp]
+theorem Rec.«end»._simpLemma_ (end1 : Option Rec) (v : Std.U32) :
```
oliver-butterley, line 40: In a couple of places like here there is escaping when not required. However this isn't critical for the functionality.

sonmarcho: Essentially escaping is needed only when there is a single identifier, right?

sonmarcho: Fixed by https://github.com/AeneasVerif/aeneas/pull/1307

PR #1305, opened 2026-08-26
Add CI to confirm ExtractBuiltinLean.ml and LeanKeywords.ml are current

oliver-butterley:
The files ExtractBuiltinLean.ml and LeanKeywords.ml are auto generated from Lean scripts. This PR adds two steps to the existing Lean ci to confirm that they aren't stale.

PR #1315, opened 2026-08-28
regenerate ExtractBuiltinLean.ml

mpenciak:
Fixes the line numbers in ExtractBuiltinLean.ml that drifted due to Iter.lean changes.

### DIFF
--- a/src/extract/ExtractBuiltinLean.ml
+++ b/src/extract/ExtractBuiltinLean.ml
@@ -37,7 +37,7 @@ let lean_builtin_types =
     mk_type "core::iter::adapters::enumerate::Enumerate"
       "core.iter.adapters.enumerate.Enumerate"
       ~kind:(KStruct [ ("iter", Some "iter"); ("count", Some "count") ]);
-    (* file: "Aeneas/Std/Core/Iter.lean", line: 892 *)
+    (* file: "Aeneas/Std/Core/Iter.lean", line: 885 *)
     mk_type "core::iter::adapters::map::Map" "core.iter.adapters.map.Map"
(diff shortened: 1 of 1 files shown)


oliver-butterley:
:thinking: does this mean that the CI introduced in #1305 failed to catch the drift?

Nadrieril:
It does seem so. Did you check whether it fired correctly while making the PR?

oliver-butterley:
> It does seem so. Did you check whether it fired correctly while making the PR?

I did check some use cases and it did what I expected. It appears I missed something... I'll investigate further.

Nadrieril:
Looks like this was likely a race condition: the PR that added the CI check stayed open for a while, and in the meantime the input changed, which was never checked twice. And because of the "diff with main" optimization, the CI jobs running against `main` didn't catch it either. I think it will work in the future

oliver-butterley:
Since #1305 was merged, none of the other PRs merged changed any of the lean files or the two derivative files and so these checks never fired in later PRs.

I think #1305 must have got merged with these files not being in sync with the Lean code. When I prepared the PR I confirmed they were correctly synced but then merging main might have led them to be out of sync. However the checks wouldn't have fired then because it would see that this particular PR didn't alter either the lean files nor the two derivative files. 

When @mpenciak is awake I'll check with him to confirm but it looks likely.

oliver-butterley:
:rofl: turns out we had a race condition with our comments! Yes, I think it will be fine in all future cases.

oliver-butterley:
:scream: we had exactly the same problem with the other derivative file. I'll do the corresponding PR to mend this.

PR #1325, opened 2026-09-01
add an `lake exe summarize` command

mpenciak:


oliver-butterley:
I'll test this extensively in our repos, just didn't get time yet. I'll do it soon!

Issue #1098, opened 2026-06-02
Name clash between variable and module


maximebuyse:
## Rust reproducer

```rust
pub mod a {
    pub fn f() {}
}

fn f(a: u8) {
    a::f()
}
```

Lean extraction:

```lean
def a.f : Result Unit := do
  ok ()

def f (a : Std.U8) : Result Unit := do
  a.f
```
Lean error:
```
Invalid field `f`: The environment does not contain `Aeneas.Std.UScalar.f`, so it is not possible to project the field `f` from an expression
  a
of type `UScalar UScalarTy.U8`
```

Here `a.f` refers to the function `f` defined in namespace `a`, but lean parses it as a projection for field `f` on the variable `a`.

will-natuvea:
We ran into this downstream while verifying Signal's SPQR, where Aeneas emits a
local `chain : chain.Chain` next to a `chain.Chain.send_key chain` call, so it
turns up on real code rather than only a corner case
(Beneficial-AI-Foundation/SparsePostQuantumRatchet-verify#101).

Minimal self-contained repro, with CI that both asserts the exact error
(`#guard_msgs`) and proves the raw form still fails to elaborate:
https://github.com/Natuvea/aeneas-1098-repro

The core of it, matching your example (Lean v4.31.0):

    namespace a
    def f : Nat := 0
    end a

    example (a : Nat) : Nat := a.f          -- error: invalid field `f` on `Nat`
    example (a : Nat) : Nat := _root_.a.f    -- elaborates

Two workarounds both resolve it: `_root_`-qualifying the call, and renaming the
bound variable (the latter is what SPQR-verify applies today, via a
post-extraction rename of `chain` to `chain'`).

If a fix in the Lean backend would be welcome, there seem to be three directions:
1. emit `_root_`-qualified names for global calls (simplest, but verbose output);
2. qualify only when the call's head segment shadows an in-scope bound variable
   (minimal output, needs the in-scope set at print time);
3. avoid binding a variable to a name equal to a top-level namespace segment.

Happy to open a PR if you have a preferred approach.

PR #1327, opened 2026-09-01
Fix 61

clementblaudeau:
Duplicate PR of : https://github.com/AeneasVerif/aeneas/pull/1326
 
Function arguments with the same name as an enclosing namespace (typically, a module or a trait) could conflict with the field notation of lean. This was reported in #1098 (and downstream in https://github.com/cryspen/hax/issues/2173, https://github.com/cryspen/aeneas/issues/61).

The fix proposed here is simple: the `collision` test also checks for conflicts with qualified names.

*AI-assisted*


clementblaudeau:
Sorry, wrong target branch. Deleting this PR

PR #1340, opened 2026-09-09
Fix problems around `Aeneas.Std.bind`

andricicezar:
Stateful effects push the Result monad to a different universe, and I had troubles with `Aeneas.Std.bind` not being properly integrated. This PR fixes some of those problems.

### DIFF
--- a/backends/lean/Aeneas/Tactic/Step/Step.lean
+++ b/backends/lean/Aeneas/Tactic/Step/Step.lean
@@ -224,6 +224,13 @@ structure Args where
   /- Syntax of the tactic provided by the user to solve the remaining proof obligations -/
   byTacSyntax : Option Syntax
 
+/-- Recognize both the monad-class bind and `Result`'s heterogeneous bind. -/
+def getBindArgs? (program : Expr) : Option (Expr × Expr) :=
+  match_expr program.consumeMData with
+  | Bind.bind _ _ _ _ m next => some (m, next)
+  | Std.bind _ _ m next => some (m, next)
+  | _ => none
+
 /-- Decompose a registered specification statement -/
 def getSpecInfoArgs (ty : Expr) : MetaM (SpecInfo × Array Expr) :=
   ty.consumeMData.withApp fun spec? args => do
@@ -260,12 +267,11 @@ def getFirstBind (goalTy : Expr) : MetaM (Bool × Expr × SpecInfo) := do
   let compTy := args[info.program_index]!
 
   trace[Step] "compTy: {compTy}"
+  trace[Step] "bind?: {compTy.consumeMData.getAppFn}"
 
-  let (bind?, args) := compTy.consumeMData.withApp (fun f args => (f, args))
-  trace[Step] "bind?: {bind?}"
-  if h: bind?.isConstOf ``bind ∧ args.size = 6
-  then pure (true, args[4], info)
-  else pure (false, compTy, info)
+  match getBindArgs? compTy with
+  | some (m, _) => pure (true, m, info)
+  | none => pure (false, compTy, info)
 
 /-- Names introduced by the `do` elaborator's `mkPatContinuation` as a
     fallback (`_xN`) when no leaf name is available — e.g. all leaves are
@@ -423,7 +429,7 @@ def getBindVarNames : TacticM (Array (Option Name)) := do
     let goalTy ← getMainTarget
     forallTelescope goalTy fun _ goalTy => do
     let m ← getSpecProgram goalTy
-    let_expr Bind.bind _ _ _ _ _ cont := m | return #[]
+    let some (_, cont) := getBindArgs? m | return #[]
     getPostNames cont
   catch _ => pure #[]
 
--- a/backends/lean/Aeneas/Tactic/Step/StepStar.lean
+++ b/backends/lean/Aeneas/Tactic/Step/StepStar.lean
@@ -387,9 +387,10 @@ def analyzeTarget : TacticM TargetKind := do
     trace[Step] "application of a registered specification statement about: {program}"
     let e ← Utils.normalizeLetBindings program
     -- Check whether this is a bind
-    if let .const ``Bind.bind .. := e.getAppFn then
-      let #[_m, _self, _α, _β, _value, cont] := e.getAppArgs
-        | throwError "Expected bind to have 6 arguments, found {← e.getAppArgs.mapM (liftM ∘ ppExpr)}"
+    if e.getAppFn.isConstOf ``Bind.bind || e.getAppFn.isConstOf ``Std.bind then
+      let arity : Nat := if e.getAppFn.isConstOf ``Bind.bind then 6 else 4
+      let some (_, cont) := Step.getBindArgs? e
+        | throwError "Expected bind to have {arity} arguments, found {← e.getAppArgs.mapM (liftM ∘ ppExpr)}"
       let names ← Step.getPostNames cont
       pure (.bind names)
     else if let .some bfInfo ← Bifurcation.Info.ofExpr e then


backends/lean/Aeneas/Std/WP.lean
```diff
@@ -146,23 +146,6 @@ theorem spec_fail_pair (e : Error) (f : α → β → Prop) :
...
-  grind
-
-theorem spec_bind {α β} {k : α -> Result β} {Pₖ : Post β} {m : Result α} {Pₘ : Post α} :
-  spec m Pₘ →
-  (forall x, Pₘ x → spec (k x) Pₖ) →
-  spec (m >>= k) Pₖ := by
-  intro Hm Hk
-  cases Hm
-  simp [Bind.bind]
-  grind only
```
sonmarcho, line 164: Why completely remove these?

backends/lean/Aeneas/Std/Primitives.lean
```diff
@@ -284,6 +284,20 @@ def Result.ofOption {a : Type u} (x : Option a) (e : Error) : Result a :=
 @[simp] theorem bind_tc_div (f : α → Result β) :
   (do let y ← div; f y) = div := by simp [bind, Bind.bind, div]
 
+/-- Associativity for `Result` computations whose value types have different universes. -/
+theorem bind_assoc_poly {α : Type u} {β : Type v} {γ : Type w}
+    (x : Result α) (f : α → Result β) (g : β → Result γ) :
+    bind (bind x f) g = bind x (fun a => bind (f a) g) := by
```
sonmarcho, line 290: I guess the current version of `bind_assoc` that is not universe polymorphic is the one that comes from the Lean standard library? I see there also exists `bind_assoc_eq`. Can't we make this one more general?

backends/lean/Aeneas/Do/Elab.lean
```diff
@@ -543,6 +556,18 @@ end
...
+def normalizeResultBinds (expr : Expr) : MetaM Expr :=
+  Meta.transform expr (post := fun e => do
+    if e.isAppOfArity ``Aeneas.Std.bind 4 then
+      let [u, v] := e.getAppFn.constLevels! | return .done e
+      if u == v then
+        let args := e.getAppArgs
+        return .done (← mkAppOptM ``Bind.bind
+          #[some (mkConst ``Aeneas.Std.Result [u]), none,
+            some args[0]!, some args[1]!, some args[2]!, some args[3]!])
+    return .done e)
```
sonmarcho, line 569: Why do we need this (the docstring is not very informative)?

backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -712,6 +731,38 @@ where
...
+      let some matcher ← matchMatcherApp? body | return .done e
+      unless matcher.discrs == #[arg] && matcher.alts.size == 1 do return .done e
+      lambdaTelescope matcher.alts[0]! fun fields body => do
+        let some (tree, count) ← tuplePatternTree? type fields 0 | return .done e
+        unless count == fields.size do return .done e
+        let cont' ← rebuildUncurryFromTree tree body
+        if cont'.containsFVar arg.fvarId! then return .done e
+        unless ← isDefEq cont cont' do
+          throwError "#decompose: tuple-pattern normalization changed the continuation"
+        return .done (mkApp2 e.appFn!.appFn! computation cont'))
```
sonmarcho, line 764: This shouldn't happen: it seems there is an issue with the do-elab?

backends/lean/Aeneas/Command/Decompose.lean
```diff
@@ -1432,14 +1483,28 @@ private def simpOnlyTarget (mvarId : MVarId) (declsToUnfold : Array Name)
...
   let unfoldNames := defNames ++ #[``_root_.Aeneas.Std.uncurry]
   let mvarId' ← mvarId.deltaTarget (unfoldNames.contains ·)
+  let hasResultBind := ((← mvarId'.getType).find?
+    (·.isAppOfArity ``Aeneas.Std.bind 4)).isSome
+  -- Keep the existing fast path for large same-universe decompositions.
+  let simpThms := if hasResultBind then
+      #[``resultBind_eq, ``resultPure_eq,
+        ``Aeneas.Std.bind_assoc_eq, ``Aeneas.Std.bind_assoc_poly,
+        ``Aeneas.Std.bind_ok, ``LawfulMonad.pure_bind]
+    else #[``Aeneas.Std.bind_assoc_eq, ``LawfulMonad.pure_bind]
```
sonmarcho, line 1507: Why an `if then else`? It seems we could use a single set of lemmas. Does it really affect performance?

backends/lean/Aeneas/Std/WP.lean
andricicezar: They are not used anywhere. There were two sets of `spec_mono` and `spec_bind`, one for `Bind.bind` and one for `Std.bind` (which is universe polymorphic), and only the latter was registered with `step`. Aeneas was applying the version for `Std.bind` everywhere.

backends/lean/Aeneas/Std/Primitives.lean
andricicezar: Yes, `bind_assoc` from the Lean standard library is not universe polymorphic. Also, `bind_assoc` is defined for `Bind.bind` and `bind_assoc_poly` is defined for `Std.bind`. Super confusing stuff. I will try to see if we can remove `Bind.bind` completely.

backends/lean/Aeneas/Do/Elab.lean
andricicezar: it was a bug. removed.

backends/lean/Aeneas/Command/Decompose.lean
andricicezar: it was a bug. removed.

backends/lean/Aeneas/Command/Decompose.lean
andricicezar: solved.

backends/lean/Aeneas/Std/Primitives.lean
sonmarcho: Yes, that's very confusing. Could we turn `bind_assoc_poly` into `Std.bind_assoc`?

sonmarcho: I see you already did this update actually...

backends/lean/Aeneas/Do/Delab.lean
```diff
@@ -112,18 +116,27 @@ where
...
+/-- Top-level `do`-block delab for `Std.bind`. -/
+@[delab app.Aeneas.Std.bind]
+meta def aeneasDelabDo : Delab := do
+  unless (← getExpr).isAppOfArity ``Aeneas.Std.bind 4 do failure
+  aeneasDelabDoCore
+
+/-- Top-level `do`-block delab for `Bind.bind`, restricted to the `Std.Result` monad
+    (the code elaborated with Lean's `do`). -/
+@[delab app.Bind.bind]
+meta def aeneasDelabBindDo : Delab := do
```
sonmarcho: The names are a bit confusing (`aeneasDelabDo` vs `aeneasDelabBindDo`).
Maybe `aeneasDelabStdBindDo` vs `aeneasDelabBindBindDo`? Or `poly` vs `mono`?

sonmarcho (commented):
Looks good, thanks! I have a single minor comment.

backends/lean/Aeneas/Do/Delab.lean
andricicezar: Renamed to `aeneasDelabStdBindDo` and `aeneasDelabBindBindDo`

sonmarcho (approved):

PR #1344, opened 2026-09-10
Add generic specifications and refactor WP around them

andricicezar:
This commit introduces generic total- and partial-correctness specifications parameterized by events and a stateful handler for those events. It also refactors the existing `spec` and `dspec` in `Std/WP.lean` to use these specifications.

Neither total nor partial correctness is defined using an inductive type. I found using a common functor (`SpecF`), then defining total correctness as the least fixed point and partial correctness as the greatest fixed point, to be the most elegant approach. Total correctness could instead be defined inductively, if there is a reason to prefer that.

One point I'm still trying to understand is the conjunctivity assumption in the admissibility theorem for partial correctness (`PartialSpec.admissible`). A handler for angelic nondeterminism is generally not conjunctive, so that theorem does not apply directly. In such a situation, we cannot use `fixpoint_induct` with partial fixpoints. We can still define programs using `partial_fixpoint` and use coinduction as an alternative way to verify them. We're not stuck on this since we do not need yet angelic nondeterminism.

The code was written entirely by AI. From what I can tell, it looks pretty standard. However, I have not previously developed total- or partial-correctness specifications for coinductive types, and I am also not familiar with CCPOs.

### DIFF
--- a/backends/lean/Aeneas/Data/Coinductive/ITree.lean
+++ b/backends/lean/Aeneas/Data/Coinductive/ITree.lean
@@ -212,6 +212,9 @@ instance : Monad (ITree.{u} E) where
   pure := ITree.ret
   bind := ITree.bind
 
+@[simp]
+theorem ITree.pure_eq_ret (r : R) : (pure r : ITree E R) = ITree.ret r := rfl
+
 @[elab_as_elim, cases_eliminator]
 def ITree.cases {E : Effect.{u}} {R}
     {motive : ITree E R → Sort v}
@@ -327,6 +330,15 @@ theorem ret_inj {E} {α} {x y} : (@ITree.ret α E x = ITree.ret y) ↔ (x = y) :
 theorem vis_inj_effect {E} {α} {e1 e2 k1 k2} : @ITree.vis α E e1 k1 = ITree.vis e2 k2
   → e1 = e2 := by grind
 
+/-- The event *and* the continuations of two equal `vis` nodes agree. -/
+theorem vis_inj {R : Type w} {E : Effect.{v}} {i i' : E.I} {k : E.O i → ITree E R}
+    {k' : E.O i' → ITree E R} (hEq : ITree.vis i k = ITree.vis i' k') :
+    i = i' ∧ HEq k k' := by
+  have hUnfold := congrArg ITree.unfold hEq
+  simp only [unfold_vis] at hUnfold
+  injection hUnfold with hEvent hCont
+  exact ⟨hEvent, hCont⟩
+
 -- Theorems to make ITree.cases compute:
 @[simp]
 theorem ITree.cases.ret {E R motive r d v x}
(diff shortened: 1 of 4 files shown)


andricicezar:
This PR is not backwards compatible with proofs that were unfolding `spec`  and  `dspec` in `ok x { Q }`. They now need to be updated to rewrite with `WP.spec_ok : ok x { Q } <-> Q x` instead

backends/lean/Aeneas/Data/Coinductive/ITree.lean
```diff
@@ -378,4 +390,119 @@ theorem ITree.div_is_bot :
...
+/-- Every element of a chain that contains a `vis` node is `div` or a `vis` node
+on the same event. -/
+theorem ITree.chain_vis_cases {c : ITree E R → Prop} (hc : chain c) {i : E.I}
+    {k' : E.O i → ITree E R} (hMem : c (ITree.vis i k')) {u : ITree E R}
+    (hu : c u) : u = ITree.div ∨ ∃ k'', u = ITree.vis i k'' := by
+  obtain hLe | hLe := hc _ _ hu hMem
+  · exact (ITree.le_vis_cases hLe).imp id fun ⟨k'', hEq, _⟩ => ⟨k'', hEq⟩
+  · exact Or.inr ((ITree.vis_le_cases hLe).imp fun k'' h => h.1)
+
+def ITree.visChain (c : ITree E R → Prop) (i : E.I) (o : E.O i) :
```
sonmarcho: A docstring would be useful

backends/lean/Aeneas/Data/Coinductive/Spec.lean
```diff
@@ -0,0 +1,509 @@
...
+
+variable {E : Effect.{v}} {α : Type u} {β : Type u'} {H : Handler.{w, v} E}
+
+/-- A postcondition on a return value and the handler's final state. -/
+abbrev HPost (H : Handler E) (α : Type u) := α → H.State → Prop
+
+/-- A predicate on an interaction tree and the handler's current state. -/
+abbrev ITreePred (H : Handler E) (α : Type u) := ITree E α → H.State → Prop
+
+def entails (P P' : α → H.State → Prop) : Prop :=
```
sonmarcho: It seems we want to use HPost here

backends/lean/Aeneas/Data/Coinductive/Spec.lean
```diff
@@ -0,0 +1,509 @@
...
+/-! ## Common functor for the least/greatest fixed point -/
+def SpecF (H : Handler E) (Q : HPost H α) (allowDivergence : Prop)
+    (X : ITreePred H α) : ITreePred H α :=
+  fun m s =>
+    ITree.cases
+      (motive := fun _ => Prop)
+      (fun value => Q value s)
+      allowDivergence
+      (fun event k => H.handle event s fun answer s' => X (k answer) s')
+      m
```
sonmarcho: Does this really give us total correctness in the case `allowDivergence` is false? I think using `ITree.cases` gives us a coinductive principle: independently of `allowDivergence` it is possible to have an infinite stream of `vis` events.

I suggest double-checking this by trying to verify something like this:
```lean
def flipCoinIgnore : Result Unit := do
  let _ <- random -- ignore the output - I just want to make sure we introduce a visible effect so that the itree doesn't evaluate to `div`
  flipCoinIgnore
partial_fixpoint
```
I think it is possible to prove a *total correctness* spec (with post-condition `_ => True`). You will need the `flipCoinIgnore.fixpoint_induct` theorem to do the proof.
This would be a reason to define the total-correctness spec as an inductive.

backends/lean/Aeneas/Data/Coinductive/Spec.lean
```diff
@@ -0,0 +1,509 @@
...
+        rw [ITree.csup_vis hc' hMem₀] at hEq
+        obtain ⟨-, hCont⟩ := vis_inj hEq.symm
+        exact eq_of_heq hCont
+      refine H.handle_mono (fun answer u' hChild => ?_) hChildren
+      exact ⟨ITree.visChain c' event answer, ITree.visChain_chain hc' event answer,
+        by rintro _ ⟨k', hMem', rfl⟩; exact hChild ⟨k', hMem'⟩, rfl⟩
+
+end Aeneas.Data.Coinductive
+
+namespace Aeneas.Data.Coinductive.StateTest
```
sonmarcho: I suggest putting the test in its own file

sonmarcho (commented):
Thanks! I essentially have one remark with regards to `TotalSpec`, which may not be correct.

What does it mean for a handler to be conjunctive? Note that it's ok if we don't support angelic nondeterminism for now: let's support one thing at a time.

backends/lean/Aeneas/Data/Coinductive/Spec.lean
andricicezar: I tried what you suggested and one cannot prove total correctness of flipCoinIgnore: https://github.com/AeneasVerif/aeneas/commit/6cee42e3bd306e49cd701fcc7ab68157e724c160#diff-aa60cd68680204ee5f45d9f9391bf90bb63a5f739ea7183d43a5eaf809cef17aR542

backends/lean/Aeneas/Data/Coinductive/Spec.lean
andricicezar: Done

backends/lean/Aeneas/Data/Coinductive/Spec.lean
andricicezar: Done

backends/lean/Aeneas/Data/Coinductive/ITree.lean
andricicezar: Done

andricicezar:
> What does it mean for a handler to be conjunctive?

The simple definition is that if `m {Q1} /\ m {Q2} ... /\ m {Qn}` then `m {fun x => Q1 x /\ Q2 x /\ .. /\  Qn x}` also has to hold.

I also merged main in to support the module system. Can you check that the `import all Init.Internal.Order.Basic` makes sense in `Spec.lean` and `Tests.lean`?

PR #1346, opened 2026-09-16
Pick the admissibility lemma by attribute

andricicezar:
The `dspec_induction` was using a admissibility lemma by name. This PR enables a user to register admissibility lemmas using `@[dspec_admissible]`, and `dspec_induction` will try the lemmas that are in this registry.


backends/lean/Aeneas/Std/WP.lean
```diff
@@ -91,6 +92,13 @@ theorem dspec_admissible {α} (p : Post α )
     constructor
 seal Result
 
+/-- The shape the `dspec_induction` tactic needs to discharge the admissibility
+side-goal it generates for a partial specification about a recursive function. -/
+@[dspec_admissible]
+theorem dspec_func_admissible {ι : Sort v} {α} (arg : ι) (p : Post α) :
```
sonmarcho: I'm a bit surprised by the choice of `ι` for the type of the argument. Why not alpha and beta for the type variables?

backends/lean/Aeneas/Tactic/Step/Tests/DspecInduction.lean
```diff
@@ -0,0 +1,186 @@
...
+
+
+def dummy_hash (_i : Std.U32) : Result Std.U32 := do
+  ok 1000#u32
+
+open ControlFlow
+
+/-- [tutorial::pseudo_random]: loop body 0:
+    Source: 'src/lib.rs', lines 258:2-260:3
+    Visibility: public -/
```
sonmarcho: I see these tests have been moved: I hadn't noticed they should be cleaned, so I'm taking the opportunity of this PR to make the relevant comments;
If we put the tests in `backends/lean` then we should consider they are hand-written and remove such comments that come from the translation.

backends/lean/Aeneas/Tactic/Step/Tests/DspecInduction.lean
```diff
@@ -0,0 +1,186 @@
...
+  simp only
+  unfold pseudo_random_loop.body
+  simp
+  by_cases ((↑x : Nat) < 100)
+  · simp [*]
+    unfold dummy_hash
+    simp
+    -- note that here, i am refraining from using the result of dummy_hash,
+    -- since its supposed to represent a hash function where we can't predict the result,
+    -- but it actually is just a constant
```
sonmarcho:
```suggestion
    -- here we treat `dummy_hash` as opaque and use `step`
```

backends/lean/Aeneas/Tactic/Step/Tests/DspecInduction.lean
```diff
@@ -0,0 +1,186 @@
...
+partial_fixpoint
+
+def second_arg_const (x y : Nat) : Result Nat :=
+  if y = 0 then .ok 0
+  else second_arg_const (x + 1) y
+partial_fixpoint
+
+-- uncomment to see the difference:
+-- #check first_arg_const.fixpoint_induct
+-- #check second_arg_const.fixpoint_induct
```
sonmarcho: It's probably better to use `#guard_msgs` to keep track of what's going on

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
```diff
@@ -19,14 +14,39 @@ import Aeneas.Tactic.Step.Step
    See the examples below to see what the tactic does
    and the manual examples show what the tactic is doing written out directly. -/
 
+
```
sonmarcho:
```suggestion
```

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
```diff
@@ -90,10 +110,21 @@ theorem curry_admissible (a1 a2 a3) (P : (a1 → a2 → a3) → Prop) [CCPO a3]
...
+    throwError "failed to prove admissibility condition: none of the theorems registered \
+                with the `dspec_admissible` attribute applies to \
+                {← instantiateMVars (← g.getType)}"
+  | thm :: thms => do
+    let s ← saveState
+    try
+      let [] ← g.apply (← mkConstWithFreshMVarLevels thm) | failure
+    catch _ =>
+      s.restore
+      applyAdmissibleThm g thms
```
sonmarcho: I think lean provides a helper that does exactly that (`saveState` then `restore`)

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
```diff
@@ -19,14 +14,39 @@ import Aeneas.Tactic.Step.Step
...
+    erase := fun declName => do
+      modifyEnv fun env =>
+        dspecAdmissibleExt.modifyState env fun s => s.erase declName
+  }
+  registerBuiltinAttribute attrImpl
+  pure attrImpl
+
+/-- Get the theorems registered with the `dspec_admissible` attribute. -/
+def getAdmissibleThms [Monad m] [MonadEnv m] : m (List Name) := do
+  pure (dspecAdmissibleExt.getState (← getEnv)).toList
```
sonmarcho: Here we're just applying all the admissibility theorems one by one without inspecting the goal: the complexity is linear in the number of theorems, which is terrible. We should do something more clever: the previous version of the code would look up exactly the theorem that applies to the function we see in the goal.

sonmarcho (commented):
Thanks for this!
I don't fully understand which problem this is trying to solve, though: in particular, the attribute seems to be used only once? Also I notice that the function that tries the admissibility theorems is linear in the number of theorems.

andricicezar:
On this branch, indeed, only one theorem is registered using `dspec_admissible`. On the separation logic branch there are two theorems, one for `dspec` and one for `idspec`. So we need a way to register multiple theorems. Do you think that having a registry is overkill?

backends/lean/Aeneas/Std/WP.lean
andricicezar: Updated to use alpha and beta

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
andricicezar: Updated the code to use a discrimination tree. I'm not sure what you mean by "the previous version of the code" since the previous version was hardcoded to apply only one lemma.

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
```diff
@@ -29,7 +25,52 @@ namespace DspecInduction
...
 open Utils
 open Lean.Order
-open Std Result
+
+/-! ## The `dspec_admissible` attribute
+
+The theorems are stored in a discrimination tree, indexed by the body of the predicate
+whose admissibility they prove: for `Order.admissible fun f => WP.dspec (f arg) post`
+the key is `WP.dspec (?f ?arg) ?post`. This allows us to look up the theorems which
+apply to a given admissibility goal, rather than trying all of them. -/
```
sonmarcho: Do we want to use `WP.dspec (?f ?arg) ?post` as the key or `WP.dspec ?e ?post` (or `WP.dspec (f ?arg) ?post`)? `WP.dspec ?e ?post` makes more sense to me than `WP.dspec (?f ?arg) ?post`

sonmarcho (commented):
I just have a single comment

backends/lean/Aeneas/Tactic/Step/DspecInduction.lean
andricicezar: Turns out that the key is already `WP.dspec ?e ?post`. I updated the comments

PR #1352, opened 2026-09-17
Extending Aeneas with first-order sequential separation logic

andricicezar:

PR #1353, opened 2026-09-17
Remove dependencies on `spec_imp_exists` and `spec_vis` in Std

andricicezar:
The way `spec_imp_exists` and `spec_vis` are stated is too syntactic, and are incompatible with adding more effects and defining the pure triples as synonyms of the effectful triples. This PR refactors some proofs that depended on these theorems, and was part of PR #1352 . The proofs look better in my opinion, so this is a positive change anyway.


andricicezar:
These proofs cannot be refactored to use `step` because they are in folder `Std`

sonmarcho (approved):
I agree this is overall cleaner: thanks! Don't hesitate to remove `spec_imp_exists` altogether if it is not used anywhere anymore.

sonmarcho (approved):

